Issue #3, opened 2024-07-25
Remove the ttf-parser dependency


rsheeter:
We don't want consumers who use skrifa to ship ttf-parser _and_ read-fonts

Look out for dodgy u32/u16 conversions when doing this

yisibl:
It's done?
https://github.com/harfbuzz/harfruzz/pull/55

PR #4, opened 2024-07-25
Use a published version of skrifa

rsheeter:
Step toward #3

Migration to read-fonts instead of skrifa left for a follow-on

Some of the gid handling is dodgy wrt 16 vs 32 bit. Fully migration to read-fonts should tidy that up.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,8 @@ unicode-bidi-mirroring = "0.3.0"
 unicode-ccc = "0.3.0"
 unicode-properties = { version = "0.1.0", default-features = false, features = ["general-category"] }
 unicode-script = "0.5.2"
-# Use git rev to handle 'true' font header
-# TODO: replace with read-fonts
-skrifa = { git = "https://github.com/googlefonts/fontations", rev = "2eb4b1c" }
+
+skrifa = "0.20"  # TODO: read-fonts instead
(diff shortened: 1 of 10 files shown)


dfrg (approved):
Was going to do this but you saved me the work :)

LaurenzV:
> # TODO: read-fonts instead

Just FYI, for 100% correctness, I'm not sure whether it will be possible to replace `skrifa` completely. There is the one method `glyph_extents`, which (among other things) relies on outlining COLR glyphs to calculate the exact extents and are used in `glyph_v_origin`, `position_mark` and `position_around_base`. I guess it's an edge case since the outlining is only necessary for COLR, but still.

behdad:
> > TODO: read-fonts instead
> 
> Just FYI, for 100% correctness, I'm not sure whether it will be possible to replace `skrifa` completely. There is the one method `glyph_extents`, which (among other things) relies on outlining COLR glyphs to calculate the exact extents and are used in `glyph_v_origin`, `position_mark` and `position_around_base`. I guess it's an edge case since the outlining is only necessary for COLR, but still.

I think in a forward-looking library, we can ignore cases where `glyph_extents` is not readily available (like CFF and COLRv1). We didn't have CFF extents for years in HB and no body complained. Also, because Windows doesn't have such a fallback mark positioning heuristic.

PR #5, opened 2024-07-25
Fix Arabic shaper bullet

rsheeter:
Attempt to address feedback in https://github.com/harfbuzz/harfruzz/pull/2#discussion_r1691552467

@khaledhosny is this version better?


behdad:
LGTM.

PR #6, opened 2024-07-26
Add contextual lookups

dfrg:
Adds the missing contextual lookups for fontations and removes all of the ttf-parser OT lookup code.

This is a huge patch but all tests pass and we're now running almost all of OT through fontations.

The remaining ttf-parser OT code revolves around shape planning: script, langsys, features, etc.


src/hb/fonta/ot/contextual.rs
```diff
@@ -235,13 +417,50 @@ impl ToU16 for BigEndian<u16> {
     }
 }
 
+fn apply_context<T: ToU16>(
+    ctx: &mut hb_ot_apply_context_t,
+    input: &[T],
+    match_func: &match_func_t,
+    lookups: &[SequenceLookupRecord],
+) -> Option<()> {
```
cmyr, line 425: 🤨

cmyr (approved):
looks like a good start!

PR #7, opened 2024-07-26
Forbid unsafe code

LaurenzV:
We already depend on `bytemuck_derive` through `font-types`, so we might as well use this to remove any remaining unsafe code.


dfrg (approved):
Yep, happy to make this change. Thanks!

rsheeter:
Nice, ty

PR #8, opened 2024-07-27
Use fontations for shape planning

dfrg:
Replaces the ttf-parser code for script, langsys and feature lookup.

GSUB/GPOS is now fully fontations.

PR #10, opened 2024-07-28
Use font_types::Tag

dfrg:
Replace `ttf_parser::Tag` with `font_types::Tag` throughout the code.

Based on #8 which should land first.


scripts/gen-tag-table.py
```diff
@@ -882,8 +882,8 @@ def rank_delta(bcp_47, ot):
 
 def hb_tag(tag):
     if tag == DEFAULT_LANGUAGE_SYSTEM:
-        return 'Tag(0)\t           '
-    return 'Tag::from_bytes(b\"%s%s%s%s\")' % tuple(('%-4s' % tag)[:4])
+        return 'Tag::new(&[0; 4])\t           '
```
khaledhosny, line 885:
```suggestion
        return 'Tag::new(&[0; 4])'
```
The two branches produce strings of the same length, the manual left padding is not needed now.

dfrg: Fixed, ty!

scripts/gen-tag-table.py
```diff
@@ -882,8 +882,8 @@ def rank_delta(bcp_47, ot):
 
 def hb_tag(tag):
     if tag == DEFAULT_LANGUAGE_SYSTEM:
-        return 'Tag(0)\t           '
-    return 'Tag::from_bytes(b\"%s%s%s%s\")' % tuple(('%-4s' % tag)[:4])
+        return 'Tag::new(&[0; 4])'
+    return 'Tag::new(b\"%s%s%s%s\")' % tuple(('%-4s' % tag)[:4])
```
cmyr, line 886: how does this work for tags that are less than 4 chars? I'm sure it does, but curious..

cmyr (approved):

scripts/gen-tag-table.py
dfrg: Not a python pro but I'm guessing the format string `"%-4s"` right pads with spaces by default.

rsheeter: REPL agrees.

```python
>>> '%-4s' % "12"
'12  '
```

rsheeter (approved):

PR #12, opened 2024-08-29
Remove deref polymorphism

simoncozens:
Deref polymorphism (using `Deref` to emulate inheritance) is a [Rust anti-pattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html). I tried to look up a definition for `set_variation`, and could not find one - this is because it was actually defined on a different object which we magically morph into; not easy to follow.

I kept a facade method for `tables` since that avoids exposing the `ttfp_face` implementation outside of the `hb_font_t` structure, but inside it, explicit is better than implicit.


simoncozens:
(Tests failing because unicode_properties is updated to the latest Unicode release.)

rsheeter (approved):
Eeek, ty

simoncozens:
Ping @drfg - `set_variation` *does not work* without this PR:

```
cannot borrow data in dereference of `harfruzz::Face<'_>` as mutable
trait `DerefMut` is required to modify through a dereference, but it is not implemented for `harfruzz::Face<'_>`
```

simoncozens:
(rebased)

PR #15, opened 2024-11-09
Be more careful when adding delta

simoncozens:
Rust has helpful convenience methods for avoiding overflows with arithmetic and avoiding casts; this prevents https://github.com/harfbuzz/rustybuzz/issues/142.

### DIFF
--- a/src/hb/ot_layout_gsubgpos.rs
+++ b/src/hb/ot_layout_gsubgpos.rs
@@ -482,7 +482,7 @@ pub(super) fn apply_lookup(
         //
         // It should be possible to construct tests for both of these cases.
 
-        end = (end as isize + delta) as _;
+        end = end.saturating_add_signed(delta);
         if end < match_positions[idx] {
             // End might end up being smaller than match_positions[idx] if the recursed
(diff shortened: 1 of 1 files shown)


alerque:
As plausible as it seemed, this isn't quite the right fix: `end` should never have been `usize` in the first place. I can submit a PR when [the dust settles](https://github.com/harfbuzz/rustybuzz/pull/146) (although I have some questions about why this fork has a separate codebase that can get out of sync and need this kind of PR instead of being a feature flag on the original port and/or being a PoC branch continuously rebased on it).

behdad:
Yeah let's keep it close to HB. If overflow happens, that shows a bug in HB code. In this case the fix is indeed to use a signed int.

behdad:
> I have some questions about why this fork has a separate codebase that can get out of sync and need this kind of PR instead of being a feature flag on the original port and/or being a PoC branch continuously rebased on it

I don't think there's a good reason for it, except that the respective owners didn't reach an agreement:

  https://github.com/googlefonts/fontations/issues/956

PR #17, opened 2025-04-11
Bring up to date with RustyBuzz main

dfrg:
This is a monster patch that brings us up to date with RustyBuzz main. 

The end result is effectively the same as a revert to the fork point, a replay of the latest RB commits, and manual reapplication of the fontations changes. Essentially, I prioritized keeping the linear RB history over maintaining my own commits.

Since the fontations changes were redone by hand, there is some churn as I moved some things around to remove unnecessary nesting and clean up some types.


rsheeter (approved):
LGTM w/o serious review, I think taking this as starting point is more useful than nitpicking it

LaurenzV:
I tried my best when porting! Seems like there have been quite a few new commits in harfbuzz since then though, especially porting the changes to AAT might be a major pain point I think. :/ IIRC this part of the codebase is the one most diverging from harfbuzz.

dfrg:
> I tried my best when porting! Seems like there have been quite a few new commits in harfbuzz since then though, especially porting the changes to AAT might be a major pain point I think. :/ IIRC this part of the codebase is the one most diverging from harfbuzz.

You did an amazing job, essentially resurrecting this project!

A general [plan](https://github.com/googlefonts/oxidize/blob/main/text/2025-03-18-harfruzz-plan.md) exists for what we intend to do moving forward. This is my primary project now so I expect things to move quickly and we will be maintaining this, including keeping parity with HarfBuzz, moving forward.

Thanks again for all of your work!

PR #21, opened 2025-05-07
Update set_digest.rs to reflect latest HarfBuzz version

behdad:
A ChatGPT collaboration.

Shows >5% speedup on Roboto shaping in HB benchmark.


behdad:
cc @LaurenzV

dfrg (approved):
Much cleaner than the templated version... and faster too! LGTM

PR #22, opened 2025-05-08
Re-export types used in the public API

dfrg:
That is `Tag` and `F2Dot14`. Looks like `FontRef` is already available.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,10 @@ extern crate alloc;
 
 mod hb;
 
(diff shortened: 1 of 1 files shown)


behdad:
LGTM, whatever that means. :D

dfrg:
Just means you can actually use all of harfruzz without having to pull in other dependencies explicitly for the types :)

behdad:
Yep. Confirmed locally.

PR #23, opened 2025-05-08
Add support for cmap format 13

dfrg:
Currently based on a git dependency so shouldn't be merged until a new version of read-fonts is released.


behdad:
Confirmed works for `hb-harfbuzz`.

PR #27, opened 2025-05-08
[gsub] Fix reverse_chain panic with NullOffset

behdad:


src/hb/ot/gsub/reverse_chain.rs
```diff
@@ -38,13 +38,21 @@ impl Apply for ReverseChainSingleSubstFormat1<'_> {
         let lookahead_coverages = self.lookahead_coverages();
 
         let f1 = |glyph: GlyphId, index| {
-            let value = backtrack_coverages.get(index as usize).unwrap();
-            value.get(glyph).is_some()
+            let value = backtrack_coverages.get(index as usize);
```
dfrg, line 41: A more Rusty approach would be: `backtrack_coverages.get(index as usize).ok().and_then(|coverage| coverage.get(glyph)).is_some()`

src/hb/ot/gsub/reverse_chain.rs
```diff
@@ -38,13 +38,21 @@ impl Apply for ReverseChainSingleSubstFormat1<'_> {
...
+                value.unwrap().get(glyph).is_some()
+            } else {
+                false
+            }
         };
 
         let f2 = |glyph: GlyphId, index| {
-            let value = lookahead_coverages.get(index as usize).unwrap();
-            value.get(glyph).is_some()
+            let value = lookahead_coverages.get(index as usize);
```
dfrg, line 50: same ^

dfrg (approved):

src/hb/ot/gsub/reverse_chain.rs
dfrg: For context, `ok()` converts a `Result` into an `Option` and `and_then` is basically a combined `map` and `flatten`.

PR #32, opened 2025-05-09
Port to HarfBuzz 10.4.0

behdad:


behdad:
mac tests failing because we need to know if `STAT` table is available. Please help:
https://github.com/googlefonts/fontations/issues/1492

src/hb/ot_shape.rs
```diff
@@ -252,8 +250,17 @@ impl<'a> hb_ot_shape_planner_t<'a> {
...
         }
 
-        // Currently we always apply trak.
-        let apply_trak = requested_tracking && self.face.aat_tables.trak.is_some();
+        // According to Ned, trak is applied by default for "modern fonts", as detected by presence of STAT table.
+        // TODO: Add STAT table check.
+        // https://github.com/googlefonts/fontations/issues/1492
+        let apply_trak = self.face.aat_tables.trak.is_some()
+            && self
+                .face
```
dfrg, line 258: This can just be `self.face.font.trak().is_ok()`. You might need to import `read_fonts::TableProvider`.

behdad: Thanks. Applied.

dfrg (approved):

PR #33, opened 2025-05-09
Update to HarfBuzz 11.2.0

behdad:


LaurenzV:
That's... A lot of new tests? 😄

behdad:
This actually brings harfRuzz up to date with HarfBuzz main as of now (604056056c8af086a4fd33855b03400bef02d48d).

behdad:
The `+57,139 −3,466` lines changed are because I enabled `emoji-clusters.tests` tests that is >3500 tests, which translate to some >55k lines of rust code.

src/hb/buffer.rs
```diff
@@ -1067,6 +1068,12 @@ impl hb_buffer_t {
         interior: Option<bool>,
         from_out_buffer: Option<bool>,
     ) {
+        if start.is_some() && end.is_some() {
+            if end.unwrap().wrapping_sub(start.unwrap()) > 255 {
```
dfrg, line 1072: The logic here is unclear to me. Looks like we bail on a range whose length is larger than 255 but only if both bounds are explicitly specified. Just below this, the bounds are potentially extended to the full buffer which could definitely be bigger than that. If this is correct, perhaps add a comment explaining why.

src/lib.rs
```diff
@@ -58,6 +58,47 @@ pub enum BufferClusterLevel {
...
     MonotoneCharacters,
     Characters,
+    Graphemes,
+}
+
+impl BufferClusterLevel {
+    #[inline]
+    fn new(level: u32) -> Self {
+        match level {
+            0 => BufferClusterLevel::MonotoneGraphemes,
```
dfrg, line 68: optional: you can replace `BufferClusterLevel::` with `Self::` for these. `Self` represents the containing type.

src/lib.rs
```diff
@@ -58,6 +58,47 @@ pub enum BufferClusterLevel {
...
+            1 => BufferClusterLevel::MonotoneCharacters,
+            2 => BufferClusterLevel::Characters,
+            3 => BufferClusterLevel::Graphemes,
+            _ => BufferClusterLevel::MonotoneGraphemes,
+        }
+    }
+    #[inline]
+    fn is_monotone(self) -> bool {
+        match self {
+            BufferClusterLevel::MonotoneGraphemes => true,
```
dfrg, line 78: Match arms are full patterns so you can union them with `|`: `Self::MonotoneGraphemes | Self::MonotoneCharacters => true,`. You can catch the remainder with `_ => false,` but keeping it explicit is fine too.

dfrg (approved):
Nice work! LGTM with one question about range logic and a couple style suggestions.

src/hb/buffer.rs
behdad: It might not be relevant to hR, as we might be always calling this function with explicit ranges. The idea in HB was that:

- Limit range to 255 to avoid O(n^2) work,

- At the same time, allow calls to buffer->unsafe_to_break() with no range, that affect the whole buffer, to succeed.

behdad: Added comment.

khaledhosny:
> The `+57,139 −3,466` lines changed are because I enabled `emoji-clusters.tests` tests that is >3500 tests, which translate to some >55k lines of rust code.

I wonder why HarfRuzz does not use something like the test runners HarfBuzz use instead of this overly verbose way.

behdad:
> > The `+57,139 −3,466` lines changed are because I enabled `emoji-clusters.tests` tests that is >3500 tests, which translate to some >55k lines of rust code.
> 
> I wonder why HarfRuzz does not use something like the test runners HarfBuzz use instead of this overly verbose way.

I assume because it easier fit into `cargo test`...

LaurenzV:
> I assume because it easier fit into cargo test

Yeah, that's the reason. It should be possible to write a custom test runner, if desired.

khaledhosny:
> > > The `+57,139 −3,466` lines changed are because I enabled `emoji-clusters.tests` tests that is >3500 tests, which translate to some >55k lines of rust code.
> > 
> > 
> > I wonder why HarfRuzz does not use something like the test runners HarfBuzz use instead of this overly verbose way.
> 
> I assume because it easier fit into `cargo test`...

Possibly. But even then, it seems to use new lines quite generously:

```
fn classdef1_multiple_002() {
    assert_eq!(
        shape(
            "tests/fonts/aots/classdef1_font3.otf",
            "\u{0011}\u{0012}\u{0013}\u{0014}\u{0015}\u{0016}\u{0017}\u{0018}\u{0019}\u{001A}\
             \u{001B}\u{001C}\u{001D}\u{001E}\u{001F}\u{0020}\u{0021}\u{0022}\u{0023}\u{0024}",
            "--features=\"test\" --no-clusters --no-glyph-names --no-positions",
        ),
        "20|\
         23|\
         24|\
         25|\
         24|\
         26|\
         27|\
         28|\
         28|\
         29|\
         30|\
         31|\
         34|\
         33|\
         34|\
         35|\
         37|\
         38|\
         38|\
         39"
    );
}
```

The output could at least be kept as single line.

behdad:
> The output could at least be kept as single line.

https://github.com/harfbuzz/harfruzz/pull/36

behdad:
> The output could at least be kept as single line.

This probably was done for better diff'ability.

PR #35, opened 2025-05-10
Skippy iter  refactor

behdad:
Some sketch on how to remove the circular cycle between skippy-iter and apply-context.

### DIFF
--- a/src/hb/aat_layout_kerx_table.rs
+++ b/src/hb/aat_layout_kerx_table.rs
@@ -156,7 +156,8 @@ fn apply_simple_kerning(
             continue;
         }
 
-        let mut iter = skipping_iterator_t::new(&ctx, i, false);
+        let mut iter = skipping_iterator_t::new(&ctx, false);
+        iter.reset(i);
 
         let mut unsafe_to = 0;
         if !iter.next(Some(&mut unsafe_to)) {
--- a/src/hb/kerning.rs
+++ b/src/hb/kerning.rs
@@ -82,7 +82,8 @@ fn machine_kern(
             continue;
         }
 
-        let mut iter = skipping_iterator_t::new(&ctx, i, false);
+        let mut iter = skipping_iterator_t::new(&ctx, false);
+        iter.reset(i);
 
         let mut unsafe_to = 0;
         if !iter.next(Some(&mut unsafe_to)) {
--- a/src/hb/ot/gpos/cursive.rs
+++ b/src/hb/ot/gpos/cursive.rs
@@ -16,7 +16,8 @@ impl Apply for CursivePosFormat1<'_> {
         let offset_data = self.offset_data();
         let entry_this = records.get(index_this)?.entry_anchor(offset_data)?.ok()?;
 
-        let mut iter = skipping_iterator_t::new(ctx, ctx.buffer.idx, false);
+        let mut iter = skipping_iterator_t::new(ctx, false);
+        iter.reset_fast(ctx.buffer.idx);
 
         let mut unsafe_from = 0;
         if !iter.prev(Some(&mut unsafe_from)) {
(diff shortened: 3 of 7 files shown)


behdad:
@dfrg @rsheeter Can I get some rust help with the last commit?

LaurenzV:
The problem is that `skippy_iterator_t` requires a reference valid for the lifetime of `'a` for buffer and face:
```rs
pub struct skipping_iterator_t<'a, 'b> {
    buffer: &'a hb_buffer_t,
    face: &'a hb_font_t<'b>,
}
```

However, what you currently have:

```rs
fn init_iters(&mut self) {
    self.iter_input = Some(skipping_iterator_t::new(self.buffer, self.face));
    self.iter_context = Some(skipping_iterator_t::new(self.buffer, self.face));
}
```

is basically equivalent to writing:

```rs
fn init_iters<'c>(&'c mut self) {
    self.iter_input = Some(skipping_iterator_t::new(self.buffer, self.face));
    self.iter_context = Some(skipping_iterator_t::new(self.buffer, self.face));
}
```

(i.e. you create a new lifetime parameter `'c` that is not in any way tied to the lifetime `'a` of `hb_ot_apply_context_t`, so this `'c` lifetime parameter is not guaranteed to live long enough, which is the requirement you need to fulfill. So in order to fix that, you would need to do:

```rs
fn init_iters(&'a mut self) {
    self.iter_input = Some(skipping_iterator_t::new(self.buffer, self.face));
    self.iter_context = Some(skipping_iterator_t::new(self.buffer, self.face));
}
``` 

so that the reference in this method is guaranteed to live as long as `'a`. However, once you change that you run into a new problem:

```
error[E0505]: cannot move out of `c` because it is borrowed
   --> src/hb/ot_layout_gsubgpos.rs:697:13
    |
669 |     impl<'a, 'b> hb_ot_apply_context_t<'a, 'b> {
    |          -- lifetime `'a` defined here
...
677 |             let mut c = Self {
    |                 ----- binding `c` declared here
...
696 |             c.init_iters();
    |             - borrow of `c` occurs here
697 |             c
    |             ^
    |             |
    |             move out of `c` occurs here
    |             returning this value requires that `c` is borrowed for `'a`
```

Now that you are saying that `init_iters` requires a mutable reference that is valid for as long as `'a`, you cannot return an owned version of `hb_ot_apply_context` in the consructor anymore! Because by the end of the constructor, `hb_ot_apply_context` is still mutable borrowed (since now you have indicated that the `&mut self` needs to be alive for the duration of the lifetime `'a`, and in Rust you cannot have a mutable reference and an owned version of the struct at the same time.

So I think the proper way to solve this is to let each client reborrow the data by using a method instead: https://github.com/LaurenzV/harfruzz/commit/87b4d5eb747636c281693c923819a260653f436f

But now the code looks similar to before again, so I'm not sure if that's what you wanted to prevent in the first place. :p (And tests seem to be failing.)

LaurenzV:
Is the part you are trying to cache the fields inside of the `matcher_t` struct?

behdad:
Thanks @LaurenzV. I vaguely understand what the problem is, and I think I can't do it the way we do it in HB. So I'm going to restructure to just cache the fileds inside `matcher_t` instead of the entire `skipping_iterator_t`. That should still give us some speedup I think.

LaurenzV:
Yep, that's probably the way!

dfrg:
+1 to this approach. In general, the solution to multiple borrow problems is to split the data into different structs that can be referenced individually.

behdad:
> +1 to this approach. In general, the solution to multiple borrow problems is to split the data into different structs that can be referenced individually.

I did that, but looks like it might not have any perf gain at this scale.

behdad:
> > +1 to this approach. In general, the solution to multiple borrow problems is to split the data into different structs that can be referenced individually.
> 
> I did that, but looks like it might not have any perf gain at this scale.

Because even if I put the matchers in the context_t, I still have to copy them into skipping_iter_t; I can't hold a reference to them since that would require a mut borrow from context_t...

With the copy, there doesn't seem to be any perf gain.

behdad:
I'm giving up on the caching / sharing. Just some refactor to bring closer to HB. @dfrg please check.

dfrg (approved):
LGTM

PR #39, opened 2025-05-11
[WIP] Cache parsed subtables per lookup

dfrg:
This avoids reparsing each subtable per glyph.


dfrg:
@behdad we may tune/change this later but I think it's worth merging for now. In addition to some decent perf gains, it also cleans up lookup dispatch and removes some unnecessary traits.

behdad:
I will study it tomorrow. Thanks.

PR #44, opened 2025-05-12
[benches] Add cold and warm versions of benchmark x hb/hr

behdad:


dfrg (approved):

behdad:
@dfrg PTAL.

benches/src/lib.rs
```diff
@@ -57,30 +101,34 @@ macro_rules! simple_bench {
...
+                let mut buffer = Some(harfruzz::UnicodeBuffer::new());
                 bencher.iter(|| {
                     test::black_box({
-                        let mut buffer = harfruzz::UnicodeBuffer::new();
-                        buffer.push_str(&text);
-                        buffer.reset_clusters();
-                        harfruzz::shape_with_plan(&shaper, &shape_plan, buffer)
+                        let mut filled_buffer = buffer.take().unwrap();
+                        filled_buffer.push_str(&text);
+                        filled_buffer.reset_clusters();
```
dfrg, line 109: Is this `reset_clusters()` call necessary? If clearing isn't doing this then that seems like a bug.

dfrg:
Other than the question about resetting clusters, buffer reuse LGTM

benches/src/lib.rs
behdad: You are right. Removing.

PR #46, opened 2025-05-12
Use kerx from fontations

dfrg:
This is currently a draft until googlefonts/fontations#1494 lands and we make a release.


behdad:
LGTM.

PR #48, opened 2025-05-12
[gsubgpos] Inline a few skippy-iter methods

behdad:
4% improvement on Roboto words HB benchmark.


dfrg (approved):
LGTM. Does `#[inline(always)]` make any difference on these?

behdad:
> LGTM. Does `#[inline(always)]` make any difference on these?

Not really. Makes it slightly worse even.

PR #49, opened 2025-05-12
Add in-house Unicode Character Database (`ucd`)

behdad:
Fixes https://github.com/harfbuzz/harfruzz/issues/29

~3% faster on Roboto HB benchmark.


behdad:
This is shaping up. I might need some guidance for the rest.

dfrg (approved):
Perf boost and removal of four dependencies, nice!

It would be great to have a section in the readme briefly describing how to regenerate the ucd.

For future work, we should simplify/unify our constant sets (GC has two representations?) and remove some of the unnecessary prefixes (`hb_gc::HB_GENERAL_CATEGORY_` is a bit of a mouthful).

PR #52, opened 2025-05-13
Enable vertical.tests and improve v_origin

behdad:
One test failing still. Working on it.

### DIFF
--- a/src/hb/glyph_metrics.rs
+++ b/src/hb/glyph_metrics.rs
@@ -1,7 +1,8 @@
+use crate::Tag;
 use read_fonts::{
     tables::{
-        glyf::Glyf, gvar::Gvar, hmtx::Hmtx, hvar::Hvar, loca::Loca, vmtx::Vmtx, vorg::Vorg,
-        vvar::Vvar,
+        glyf::Glyf, gvar::Gvar, hmtx::Hmtx, hvar::Hvar, loca::Loca, mvar::Mvar, vmtx::Vmtx,
+        vorg::Vorg, vvar::Vvar,
     },
     types::{BoundingBox, F2Dot14, Fixed, GlyphId, Point},
     FontRef, TableProvider,
@@ -15,6 +16,7 @@ pub(crate) struct GlyphMetrics<'a> {
     vvar: Option<Vvar<'a>>,
     vorg: Option<Vorg<'a>>,
     glyf: Option<GlyfTables<'a>>,
+    mvar: Option<Mvar<'a>>,
     ascent: i16,
     descent: i16,
 }
@@ -42,6 +44,7 @@ impl<'a> GlyphMetrics<'a> {
         } else {
             None
         };
+        let mvar = font.mvar().ok();
         let (ascent, descent) = if let Ok(os2) = font.os2() {
             (os2.s_typo_ascender(), os2.s_typo_descender())
         } else if let Ok(hhea) = font.hhea() {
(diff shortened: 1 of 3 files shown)


behdad:
Okay I see what's going on. We're in the codepath where the font is ttf and has no vmtx. So we try to calculate the v_origin using glyph extents. In HB, if we're in glyf and coords are set, we calculate the glyph bounds from the outline. I believe this is not implemented in hR?

src/hb/glyph_metrics.rs
```diff
@@ -126,16 +130,14 @@ impl<'a> GlyphMetrics<'a> {
         let gid = gid.into();
         let mut bearing = if let Some(vmtx) = self.vmtx.as_ref() {
             vmtx.side_bearing(gid).unwrap_or_default() as i32
-        } else if let Some(extents) = self.extents(gid, coords) {
-            return Some(extents.y_min);
```
behdad, line 130: Side bearing is not related related to extents.

src/hb/glyph_metrics.rs
```diff
@@ -126,16 +130,14 @@ impl<'a> GlyphMetrics<'a> {
...
         } else {
             return None;
         };
         if !coords.is_empty() {
             if let Some(vvar) = self.vvar.as_ref() {
                 bearing += vvar.tsb_delta(gid, coords).unwrap_or_default().to_i32();
             } else if let Some(deltas) = self.phantom_deltas(gid, coords) {
-                bearing += deltas[2].x.to_i32();
+                bearing += deltas[3].y.to_i32();
```
behdad, line 140: I think we want y-max delta, right?

src/hb/glyph_metrics.rs
```diff
@@ -147,20 +149,66 @@ impl<'a> GlyphMetrics<'a> {
...
+                        .round()
+                        .to_i32();
                 }
             }
             origin
         } else if let Some(extents) = self.extents(gid, coords) {
-            if self.vmtx.is_some() {
-                extents.y_min + self.top_side_bearing(gid, coords).unwrap_or_default()
+            let origin = if self.vmtx.is_some() {
+                let mut origin = Some(extents.y_max);
```
behdad, line 162: Note how this used to be `y_min`. But `y_max` is intended.

src/hb/glyph_metrics.rs
dfrg: oops, yes. Copy pasta error :)

src/hb/glyph_metrics.rs
```diff
@@ -147,20 +149,66 @@ impl<'a> GlyphMetrics<'a> {
             let mut origin = vorg.vertical_origin_y(gid) as i32;
             if !coords.is_empty() {
                 if let Some(vvar) = self.vvar.as_ref() {
-                    origin += vvar.v_org_delta(gid, coords).unwrap_or_default().to_i32();
+                    origin += vvar
+                        .v_org_delta(gid, coords)
+                        .unwrap_or_default()
+                        .round()
```
dfrg, line 155: `Fixed::to_i32()` should already round: https://github.com/googlefonts/fontations/blob/6449c344a633da2b8ad10ab4dc6d680f561eca86/font-types/src/fixed.rs#L309

dfrg (approved):
LGTM. TIL that missing vmtx means that vertical phantom points cannot be trusted. Thanks!

dfrg:
> Okay I see what's going on. We're in the codepath where the font is ttf and has no vmtx. So we try to calculate the v_origin using glyph extents. In HB, if we're in glyf and coords are set, we calculate the glyph bounds from the outline. I believe this is not implemented in hR?

Right. We don’t have any mechanism to load a glyf outline in read-fonts. It’s not impossible but would require some substantial changes if we want to support variations.

PR #53, opened 2025-05-14
Add hb_cache_t and use for cmap

behdad:
ChatGPT collaboration.

### DIFF
--- a/src/hb/cache.rs
+++ b/src/hb/cache.rs
@@ -1,60 +1,56 @@
-use core::marker::PhantomData;
 use core::sync::atomic::{AtomicU16, AtomicU32, Ordering};
 
 /// Trait for atomics used in cache storage
-pub trait AtomicStorage: Default {
+pub trait AtomicStorage: Sized {
     const BITS: usize;
     fn get(&self) -> u32;
     fn set(&self, val: u32);
+    fn default() -> Self;
 }
 
-/// AtomicU16 wrapper
-#[derive(Debug)]
-pub struct Atomic16(AtomicU16);
-impl Default for Atomic16 {
-    fn default() -> Self {
-        Self(AtomicU16::new(u16::MAX))
-    }
-}
-impl AtomicStorage for Atomic16 {
+/// Implement AtomicStorage directly for AtomicU16
+impl AtomicStorage for AtomicU16 {
     const BITS: usize = 16;
 
     fn get(&self) -> u32 {
-        self.0.load(Ordering::Relaxed) as u32
+        self.load(Ordering::Relaxed) as u32
     }
+
     fn set(&self, val: u32) {
-        self.0.store(val as u16, Ordering::Relaxed)
+        self.store(val as u16, Ordering::Relaxed)
+    }
+
+    fn default() -> Self {
+        Self::new(u16::MAX)
     }
 }
 
-/// AtomicU32 wrapper
-#[derive(Debug)]
-pub struct Atomic32(AtomicU32);
-impl Default for Atomic32 {
-    fn default() -> Self {
-        Self(AtomicU32::new(u32::MAX))
-    }
-}
-impl AtomicStorage for Atomic32 {
+/// Implement AtomicStorage directly for AtomicU32
+impl AtomicStorage for AtomicU32 {
     const BITS: usize = 32;
 
     fn get(&self) -> u32 {
-        self.0.load(Ordering::Relaxed)
+        self.load(Ordering::Relaxed)
     }
+
     fn set(&self, val: u32) {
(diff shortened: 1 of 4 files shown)


src/hb/cache.rs
```diff
@@ -0,0 +1,133 @@
...
+            "KEY_BITS must be >= log2(CACHE_SIZE)"
+        );
+
+        debug_assert!(
+            KEY_BITS + VALUE_BITS <= cache_bits + T::BITS,
+            "KEY_BITS + VALUE_BITS must fit in CACHE_BITS + T::BITS"
+        );
+
+        Self {
+            values: std::array::from_fn(|_| T::default()),
```
dfrg, line 97: Should we initialize to INVALID rather than default() (which would be zero)?

behdad: Right. ChatGPT...

behdad:
@dfrg what should we do with this?

src/hb/cache.rs
```diff
@@ -0,0 +1,139 @@
+use core::marker::PhantomData;
+use core::sync::atomic::{AtomicU16, AtomicU32, Ordering};
+
+/// Trait for atomics used in cache storage
+pub trait AtomicStorage: Default {
```
dfrg, line 5: If you add a `default` (or maybe `initial`) method to this trait, you should be able to avoid the wrappers and impl it directly on the core types.

src/hb/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+        Self {
+            values: core::array::from_fn(|_| T::default()),
+            _marker: PhantomData,
+        }
+    }
+
+    #[inline]
+    pub fn get(&self, key: u32) -> Option<u32> {
+        let index = (key as usize) & (CACHE_SIZE - 1);
+        let stored = T::get(&self.values[index]);
```
dfrg, line 117: I think this can just be `self.values[index].get()`. `T` is constrained to `AtomicStorage` so the method is available.

src/hb/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    #[inline]
+    pub fn set(&self, key: u32, value: u32) -> bool {
+        if (key >> KEY_BITS) != 0 || (value >> VALUE_BITS) != 0 {
+            return false;
+        }
+
+        let index = (key as usize) & (CACHE_SIZE - 1);
+        let packed = ((key >> (CACHE_SIZE as u32).ilog2()) << VALUE_BITS) | value;
+        T::set(&self.values[index], packed);
```
dfrg, line 136: Same

src/hb/charmap.rs
```diff
@@ -42,12 +46,17 @@ impl<'a> Charmap<'a> {
...
-        Self::default()
+        Self {
+            subtable: None,
+            vs_subtable: None,
+            cache,
+        }
     }
 
-    pub fn map(&self, mut c: u32) -> Option<GlyphId> {
+    fn _map(&self, mut c: u32) -> Option<GlyphId> {
```
dfrg, line 59: In Rust, the leading underscore generally implies something that is unused rather than private. We typically use the suffixes `_impl` or `_inner` when we’re delegating work to an internal function.

dfrg (approved):
Apologies for the delay. The logic LGTM with a few suggestions on style.

I would like to avoid adding more C style type names moving forward. It adds a lot of noise and makes the code more difficult to read for a Rust developer. I plan on doing some passes to clean this up soon so feel free to land this PR without those changes.

src/hb/cache.rs
behdad: I'm not sure. Setting the default for the core types to set to -1 sounds phishy to me. :)

behdad: It feels to me like I don't know what I'm talking about :)). Yeah I think I agree with you. Will try removing.

behdad:
@dfrg PTAL.

behdad:
Oh it all broke.

src/hb/cache.rs
behdad: I'm not sure how to do this.

nicoburns: I think the idea is to add `fn default() -> Self` directly to the `AtomicStorage` trait (allowing you to remove the `: Default` bound on `AtomicStorage`). You can then use `AtomicStorage::default()` instead of `Default::default()` when initialising the type. You cannot implement `Default` for core types (in general, you cannot implement a foreign trait on a foreign type)

behdad: Thanks. Done. PTAL.

dfrg (approved):
One nit about unnecessary `PhantomData` otherwise this is good to go.

PR #57, opened 2025-06-04
Some API changes

dfrg:
This is a set of changes designed for three purposes: 1) add types where we can hang various caches 2) avoid lifetimes associated with cached data, and 3) make the API a bit more Rusty.

The general design results in four types:

* `ShaperData`: this contains preprocessed data that is invariant to size and variations. Lookup/subtable metadata, cmap cache, etc. You would create one of these per font.
* `ShaperInstance`: data that is dependent on variations. This currently just contains the normalized coordinates but should ultimately include things like precomputed IVS scalars. One per font and location in variation space.
* `Shaper`: an immutable transient type that combines references to values of the above two types with a `FontRef` and provides the entry point to shaping.
* `ShaperBuilder`: used to construct a `Shaper` allowing for future extension of configuration options without breaking changes or long parameter lists.
    * A future extension here might be support for overriding glyph metrics

In addition, the `shape` and `shape_with_plan` functions become methods on `Shaper` and the parameter order is changed to match HB.

Usage now looks like the following:
```rust
// Read some font data
let font_data = std::fs::read("/path/to/some/font.ttf").unwrap();
// Create a FontRef
let font = FontRef::from_index(&font_data, 0).unwrap();

// Note: the following two types have no lifetimes and can be cached

// Create the store for our invariant cached data (lookups, digests, etc)
let shaper_data = ShaperData::new(&font);

// Create the store for our variation specific cached data (coords, IVS scalars, etc)
let shaper_instance = ShaperInstance::from_variations(&font, [("wght", 550.0), ("wdth", 75.5)]);
// .. or from normalized coordinates
let shaper_instance = ShaperInstance::from_coords(&font, [-0.9, 0.4].map(NormalizedCoord::from_f32));
// .. or from a named instance index
let shaper_instance = ShaperInstance::from_named_instance(&font, 2);

// Now build a Shaper which is a transient type combining the above objects
// and some configuration
let shaper = shaper_data
    // Initialize the builder for the given font
    .shaper(&font)
    // Set the instance
    .instance(Some(&shaper_instance))
    // Set the desired point size for trak
    .point_size(Some(12.0))
    // Build the shaper
    .build();

// Make a buffer
let mut buffer = UnicodeBuffer:new();
buffer.push_str("hello world");
buffer.guess_segment_properties();

// And shape!
let glyph_buffer = shaper.shape(buffer, &[]);
```

This design is meant to make the costs explicit and allow the user to cache as desired. Future work could build on this to provide an additional type (`ShaperFont`?) that acts more like HB and manages caching of instances and shape plans internally.


dfrg:
@behdad @LaurenzV your opinions on this would be appreciated!

behdad:
Looks good. `shape_with_plan` will still need to be changed, because it should also receive the features again. Maybe do that so we don't break the API again later?

I'll see if I can update hb-harfruzz to these changes soon.

LaurenzV:
I'll probably have to actually try it somewhere (e.g. in resvg) to be able to judge better, but overall it seems like a good change. :)

benches/src/lib.rs
```diff
@@ -45,20 +45,15 @@ macro_rules! simple_bench {
...
-                        }
-                        let shaper = shaper_font.shaper(&font, &coords);
+                        let instance = harfruzz::ShaperInstance::from_variations(&font, vars.iter().map(|var| (var.tag, var.value)));
+                        let shaper = data.shaper(&font).instance(Some(&instance)).build();
                         let mut buffer = harfruzz::UnicodeBuffer::new();
                         buffer.push_str(&text);
                         buffer.guess_segment_properties();
                         let shape_plan = harfruzz::ShapePlan::new(&shaper, buffer.direction(), Some(buffer.script()), buffer.language().as_ref(), &[]);
-                        harfruzz::shape_with_plan(&shaper, &shape_plan, buffer)
+                        shaper.shape_with_plan(buffer, &shape_plan)
```
behdad, line 56: These don't need the empty features slice?

src/hb/face.rs
```diff
@@ -65,37 +217,85 @@ pub struct hb_font_t<'a> {
...
-    /// Used for optical-sizing in Apple fonts.
+    /// # Panics
     ///
-    /// `None` by default.
+    /// Will panic when debugging assertions are enabled if the buffer and plan have mismatched
+    /// properties.
+    pub fn shape_with_plan(
+        &self,
+        buffer: UnicodeBuffer,
+        plan: &hb_ot_shape_plan_t,
```
behdad, line 267: I wonder if the plan should come either before buffer, or after features. In the middle kinda doesn't appeal to me.

behdad:
HarfBuzz patch in https://github.com/harfbuzz/harfbuzz/pull/5355

Seems to work as before, except that I'm seeing `test-draw` fail, which I do not expect. Investigating.

src/hb/ot_shape.rs
```diff
@@ -504,7 +505,7 @@ fn setup_masks(ctx: &mut hb_ot_shape_context_t) {
         func(ctx.plan, ctx.face, ctx.buffer);
     }
 
-    for feature in &ctx.plan.user_features {
+    for feature in ctx.features {
```
dfrg, line 508: @behdad this is the only place where I use the features provided by `shape_with_plan`. I'm not sure if this matches HB

src/hb/face.rs
```diff
@@ -65,37 +216,85 @@ pub struct hb_font_t<'a> {
...
     ///
-    /// `None` by default.
-    #[inline]
-    pub fn set_pixels_per_em(&mut self, ppem: Option<(u16, u16)>) {
-        self.pixels_per_em = ppem;
+    /// If you plan to shape multiple strings, prefer [`shape_with_plan`](Self::shape_with_plan).
+    /// This is because [`ShapePlan`](crate::ShapePlan) initialization is pretty slow and should preferably
+    /// be called once for each shaping configuration.
+    pub fn shape(&self, mut buffer: UnicodeBuffer, features: &[Feature]) -> GlyphBuffer {
+        buffer.0.guess_segment_properties();
```
behdad, line 240: As suggested before, let's remove this.

src/hb/face.rs
```diff
@@ -65,37 +216,85 @@ pub struct hb_font_t<'a> {
...
+    /// Will panic when debugging assertions are enabled if the buffer and plan have mismatched
+    /// properties.
+    pub fn shape_with_plan(
+        &self,
+        plan: &ShapePlan,
+        buffer: UnicodeBuffer,
+        features: &[Feature],
+    ) -> GlyphBuffer {
+        let mut buffer = buffer.0;
+        buffer.guess_segment_properties();
```
behdad, line 270: Same here.

behdad:
LGTM. Left one minor comment about removing `guess_segment_properties` call by default.

rsheeter:
Suggestions:

* `s/ShaperData/ShaperCache` to clarify it's role
* The transient Shaper doesn't feel like it's pulling it's weight
   * What if ShaperInstance was called Shaper and you called shape on that with a thing you build, say ShapeRequest?
* The input and output of shape could be structs so we can add to them if needed?

Sketch:

```rust
// A FontRef is something everyone needs
let font_data = std::fs::read("/path/to/some/font.ttf").unwrap();
let font = FontRef::from_index(&font_data, 0).unwrap();

// Note: the following two types have no lifetimes and can be cached

// Create the store for our invariant cached data (lookups, digests, etc)
let cache = ShaperCache::new(&font);

// Create the store for our variation specific cached data (coords, IVS scalars, etc)
let shaper = Shaper::from_variations(&font, [("wght", 550.0), ("wdth", 75.5)]);
// .. or from normalized coordinates
let shaper = Shaper::from_coords(&font, [-0.9, 0.4].map(NormalizedCoord::from_f32));
// .. or from a named instance index
let shaper = Shaper::from_named_instance(&font, 2);

// Make a reusable buffer
let mut buffer = UnicodeBuffer:new();
buffer.push_str("hello world");

// Now we can submit ShapeRequest to our Shaper
// Surrender ownership of our buffer to the request
let request = ShapeRequest::new(buffer)
   .with_cache(&cache)  // maybe you *must* provide a cache?
   .with_size(Some(12.0))
  .with_features(...)
  .build();

// And shape! Which explicitly returns our UnicodeBuffer to us. Possibly wrapped in a container so we can return other data without breaking people in future?
let (result, our_buffer_ready_to_reuse) = shaper.shape(request);
```

behdad:
Speaking of names, looks like `ShaperData` is what `hb_face_t` is, and `ShaperInstance` is `hb_font_t`. I wonder if it would be less confusing to name them as such. No pressure. :)

PR #62, opened 2025-06-09
[buffer] Fix buffer size enlargement

behdad:
We were wrongly, comparing with self.len, not the self.info.len which is the current length of the info/pos arrays. This was resulting in inadverently shrinking the buffer sometimes, causing out-of-bounds access and crash, for example in text-rendering-tests MORX-31 test.


dfrg (approved):
I'm surprised this didn't cause more issues. Fix LGTM!

PR #67, opened 2025-06-10
Create a CHANGELOG

DJMcNab:
This format is loosely based on the Linebender standards, but also follows patterns laid out by the prior changelog.

Keeping a CHANGELOG is important for your users to keep track of what changed, especially why they should upgrade.

An alternative would be to keep the old changelog contents around; there are reasonable arguments for that, and I would be happy to port the content in.
The arguments against are that the old version numbers don't make sense/conflict, among other things.

I notice that your git tags don't start with a `v`; the ones in fontations do so, but HarfBuzz and RustyBuzz do not do so.


behdad:
Thank you!

PR #73, opened 2025-06-25
chore: clippy fixes

nyurik:
Enable Pedantic clippy lints, while also allowing all the ones that got triggered.  Not all of these lints should be fixed, but this gives a list of things to consider fixing in subsequent PRs.

The list was generated with:

```sh
cargo clippy --all-targets --workspace --all-features --message-format=json --quiet | jq -r '.message.code.code | select(. != null and startswith("clippy::"))' | sort | uniq -c | sort -h -r
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,62 @@ std = ["read-fonts/std"]
 [dev-dependencies]
 pico-args = { version = "0.5", features = ["eq-separator"] }
 libc = "0.2"
+
+[lints.rust]
+# TODO: enable this and fix related code
+unused_qualifications = "allow"
+
+[lints.clippy]
+cargo = { level = "warn", priority = -1 }
+pedantic = { level = "warn", priority = -1 }
+# These are the currently triggering clippy lints, which should be fixed in the subsequent PRs.
+# This list was generated with this command:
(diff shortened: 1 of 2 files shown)


behdad:
Thanks. Please fix this rustfmt issue:

  https://github.com/harfbuzz/harfrust/actions/runs/15886672730/job/44801568528?pr=73

Also, maybe we can turn the github clippy job to fail now with these.

nyurik:
Incidentally, a lot of projects nowadays enable [pre-commit ci](https://pre-commit.com/) (not to be confused with GIT pre-commit, which runs locally).  The idea is simple - it connects to a public github repo, and whenever there is a pull request, it does several well defined checks, and tries to do some basic fixes like `cargo fmt`, trailing spaces, last newline in a file, etc - and if fixed, it pushes another commit to the same repo.  This massively simplifies contributions by the users - because now instead of CI being a blocker (requiring me to push another commit with a simple `cargo fmt` fix, and for the reviewer to click the approve button to run the CI, the fixed commit gets pushed automatically - keeping the code clean and not annoying casual contributors.  See [my example](https://github.com/nyurik/automotive_diag/blob/main/.pre-commit-config.yaml) that I used in many projects.  Simply add that file to the root, and enable it at https://pre-commit.ci/

PR #74, opened 2025-06-25
chore: a few clippy fixes

nyurik:
this is based on #73, fixing some of the found issues. Once that PR is merged, it can be used to keep track of which are fixed and which remain


dfrg:
This LGTM.

Would you like to update the lint config to account for the fixes in this PR or defer that to a follow up?

nyurik:
I can do it in a follow up

PR #79, opened 2025-06-26
chore: fix redundant_closure_for_method_calls

nyurik:
TBD: I am not sure this is a good lint here - perhaps disable it instead?


nyurik:
it's not bad - the only goal is readability. I have been using this pattern, but some might object to it.  TBH, i prefer to follow most clippy lints unless they really get in the way :)

dfrg:
Apologies for the delay. This LGTM and I'd be happy to merge once the conflict is resolved.

nyurik:
@dfrg thx, merged from main.  Please make sure to use `squash and merge` when merging (there is actually a setting for that to keep the history clean) - on the settings general page:

image

PR #83, opened 2025-06-26
chore: migrate to justfile, and make CI use it

nyurik:
* improve CI workflow:
  * move all project-specific CI logic into the `justfile` - this allows developers to re-run the same steps as CI locally, without the need to replicate whatever is being done in the CI
  * add test, test-msrv, coverage, and the release jobs
* implement many basic commands in just, e.g. `just coverage` will open a local coverage report with all tests
* automatically treat all warnings as errors when running in CI, but the same commands will only show warnings when running locally - allowing quicker development while keeping checked-in code clean


nyurik:
I simplified this PR, moving dependabot updates into #120 , and release-plz will be a subsequent separate PR to keep things simpler

behdad:

PR #89, opened 2025-07-23
chore: remove Cargo.lock

nyurik:
Libraries should not contain the lock file, as that does not change anything for the users of the library, but may hide bugs for the maintainers of the library itself.

P.S. Note that this used to be part of #83


nyurik:
@waywardmonkeys this is true - but is there a specific reason you want to keep the lock?  Having lock means CI has to also test removing it and doing a build without it.

dfrg:
We have (and intend to retain) a very small dependency graph, almost all of which we control. Given that and the desire to keep this mostly aligned with the fontations repo, I’m in favor of removing the lock file. We can restore it later if it becomes a problem.

PR #101, opened 2025-08-01
Context rules abstraction; try HB fast paths

dfrg:
Chained contextual rules still TODO


behdad:
Nice: NotoNastaliqUrdu from 3.34x HB down to 2.7x HB

PR #108, opened 2025-08-02
More efficient subtable rematerialization

dfrg:
Flattens the subtable format into an enum so we're doing an exhaustive match instead of a comparison chain on (bool, u8). Also removes an unnecessary indirection through the LookupCache.


behdad:
This looks like it slows down Roboto shaping by 10% in my measurement.

dfrg:
Thanks. I think no measurable difference is acceptable because the structure of the code is a bit cleaner and this will eventually allow us to remove the secondary subtable cache entirely once we have faster read-fonts parsing— the code in #109 doesn’t use it.

behdad:
LGTM.

dfrg:
This was superseded by #133

PR #109, opened 2025-08-02
Experiment: check overhead of read-fonts

dfrg:
This does some ugly hand rolled parsing of the ligature table. No unsafe, so this is meant to measure the straight overhead of read-fonts style parsing for lookup subtables.

Based on #108 but this one should not be merged. For measurement only.

### DIFF
--- a/src/hb/ot/contextual.rs
+++ b/src/hb/ot/contextual.rs
@@ -1,15 +1,17 @@
 use super::{coverage_index, covered, glyph_class};
 use crate::hb::ot_layout_gsubgpos::OT::hb_ot_apply_context_t;
 use crate::hb::ot_layout_gsubgpos::{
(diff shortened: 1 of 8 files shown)


behdad:
15% improvement in `Roboto-Regular.ttf/en-thelittleprince.txt` over #108

behdad:
> 15% improvement in `Roboto-Regular.ttf/en-thelittleprince.txt` over #108

I consider 15% very significant. Note that we're currently 2.5% slower than HB. So, a 15% improvement here is, like, over one-third of the HB shaping time.

dfrg:
Done for PairPos subtables as well.

dfrg:
Keeping this one open for now because it still demonstrates some major gains we could achieve with optimized liga and kern implementations.

dfrg:
Closing. New state tracked in #136

PR #114, opened 2025-08-06
Check coverage before reading subtables

dfrg:
The intent is to avoid both redundant and unnecessary work. For each subtable, before doing the full parse, read the primary coverage table and then compute the index for the current glyph. If successful, bundle up the glyph, coverage table and coverage index so we can pass that data on to the subtable apply function. If unsuccessful, don't parse the subtable at all.


behdad:
I'm fine with this going on. But if we can do without it (speeding up fontations reading), I prefer that, to remain closer to HB.

We tried this approach in HB before but it didn't pay off, since we could, instead, just cache the subtable apply function and use that as a function pointer, which is what we do.

PR #117, opened 2025-08-08
Adapt buffer var allocation bookkeeping from HB

behdad:
Necessary for porting some of the caches.

### DIFF
--- a/src/hb/buffer.rs
+++ b/src/hb/buffer.rs
@@ -188,6 +188,7 @@ macro_rules! declare_buffer_var {
         };
 
         #[inline]
+        #[allow(dead_code)]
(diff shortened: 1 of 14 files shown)


behdad:
I tried minimizing duplication in the macro parameters to the extent possible without the `parse` crate.

dfrg:
I have some reservations about this in general but I’m fine with landing this in its current state so we can move forward with other optimizations.

behdad:
> I have some reservations about this in general but I’m fine with landing this in its current state so we can move forward with other optimizations.

Thanks. Happy to adjust later. I went this way to avoid having to change all callsites. But the macro is easy enough to tinker with.

PR #119, opened 2025-08-08
chore(example): use cstr .asptr()

nyurik:
`CStr` are so much cleaner...

### DIFF
--- a/examples/shape.rs
+++ b/examples/shape.rs
@@ -281,7 +281,7 @@ fn parse_cluster(s: &str) -> Result<harfrust::BufferClusterLevel, String> {
 
 fn system_language() -> harfrust::Language {
     unsafe {
-        libc::setlocale(libc::LC_ALL, b"\0" as *const _ as *const i8);
+        libc::setlocale(libc::LC_ALL, c"".as_ptr());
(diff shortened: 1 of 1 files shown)


behdad:
I just saw your MSRV question on fontations. Does this PR also require higher MSRV than the declared 1.75? I understand it's in the example code, but I think it makes sense to stick with MSRV declared in our Cargo.toml.

Sorry if I'm completely off. I'm new to the ecosystem.

nyurik:
I just realized it as well - my other PR to bring CI up to date has caught it too -- https://github.com/harfbuzz/harfrust/actions/runs/16838706781/job/47704401185?pr=83

nyurik:
We may want to revert this until MSRV is 1.77+ i guess - i can do another PR for that

behdad:
No worries. I revert it locally.

nyurik:
see, that's why we need to improve CI to catch these things automatically :D

behdad:
Fontations jumped MSRV to 1.80, so doing same here now.

behdad:
> Fontations jumped MSRV to 1.80, so doing same here now.

I'm reapplying this.

Issue #118, opened 2025-08-08
`would_apply` logic error


behdad:
It seems like it only tries the first subtable. I think this is intended:
```diff
diff --git a/src/hb/ot/lookup.rs b/src/hb/ot/lookup.rs
index 363fa2f..8fe9110 100644
--- a/src/hb/ot/lookup.rs
+++ b/src/hb/ot/lookup.rs
@@ -370,7 +370,9 @@ impl LookupInfo {
                 Subtable::ChainedContextFormat3(subtable) => subtable.would_apply(ctx),
                 _ => false,
             };
-            return Some(result);
+            if result {
+                return Some(result);
+            }
         }
         None
     }
```
I honestly don't understand why an `Option<bool>` is returned, not a bool.

PR #121, opened 2025-08-08
[would_apply] Fix subtable logic

behdad:
We should check all subtables until one returns true.

Fixes https://github.com/harfbuzz/harfrust/issues/118


behdad:
Well needs building a custom Indic font. I might be able to do that.

behdad:
> Seems testable?

Added.

PR #123, opened 2025-08-09
Pass `&mut hb_glyph_info_t` to match functions

behdad:
Going with `mut` was fighting a losing battle with the borrow checker. I made it to passing `&hb_glyph_info_t`. My plan is to use `UnsafeCell` for free vars in `hb_glyph_info_t` so we can modify them without mutability.

### DIFF
--- a/src/hb/ot/contextual.rs
+++ b/src/hb/ot/contextual.rs
@@ -1,4 +1,5 @@
 use super::{coverage_index, covered, glyph_class};
+use crate::hb::buffer::hb_glyph_info_t;
 use crate::hb::ot_layout_gsubgpos::OT::hb_ot_apply_context_t;
 use crate::hb::ot_layout_gsubgpos::{
     apply_lookup, match_backtrack, match_glyph, match_input, match_lookahead, may_skip_t,
@@ -29,7 +30,11 @@ impl WouldApply for SequenceContextFormat1<'_> {
                         let input = rule.input_sequence();
                         ctx.glyphs.len() == input.len() + 1
                             && input.iter().enumerate().all(|(i, value)| {
-                                match_glyph(ctx.glyphs[i + 1], value.get().to_u16())
+                                let info = hb_glyph_info_t {
+                                    glyph_id: ctx.glyphs[i + 1].into(),
+                                    ..hb_glyph_info_t::default()
+                                };
+                                match_glyph(&info, value.get().to_u16())
                             })
                     })
                     .unwrap_or(false)
@@ -62,10 +67,13 @@ impl WouldApply for SequenceContextFormat2<'_> {
                     rule.map(|rule| {
                         let input = rule.input_sequence();
                         ctx.glyphs.len() == input.len() + 1
-                            && input
-                                .iter()
-                                .enumerate()
-                                .all(|(i, value)| match_fn(ctx.glyphs[i + 1], value.get()))
+                            && input.iter().enumerate().all(|(i, value)| {
+                                let info = hb_glyph_info_t {
+                                    glyph_id: ctx.glyphs[i + 1].into(),
+                                    ..hb_glyph_info_t::default()
+                                };
+                                match_fn(&info, value.get())
+                            })
                     })
                     .unwrap_or(false)
                 })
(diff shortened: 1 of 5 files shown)


behdad:
> I'm out of ideas how to implement the caches that use the `hb_glyph_info_t` vars slots, in safe rust.

I can get something like this with unsafe, if we choose to allow that. Maybe I prototype and we see the speedup? @rsheeter

```rust
    fn syllable_mut(&self) -> &mut u8 { 
        let shape = hb_glyph_info_t::SYLLABLE_VAR; 
        let start = shape.start(); 
        let count = shape.count(); 
        debug_assert_eq!(count, 1); 
        unsafe { 
            let base_const: *const u8 = core::ptr::addr_of!(self.vars).cast(); 
            let p: *mut u8 = base_const.add(start as usize) as *mut u8; 
            &mut *p 
        } 
    } 
```

behdad:
Thanks to Chad's work in https://github.com/harfbuzz/harfrust/pull/124 I managed to pass `&mut hb_glyph_info_t` to match functions here. So, no `unsafe` needed it seems. Yay.

src/hb/ot/contextual.rs
```diff
@@ -29,7 +30,11 @@ impl WouldApply for SequenceContextFormat1<'_> {
                         let input = rule.input_sequence();
                         ctx.glyphs.len() == input.len() + 1
                             && input.iter().enumerate().all(|(i, value)| {
-                                match_glyph(ctx.glyphs[i + 1], value.get().to_u16())
+                                let mut info = hb_glyph_info_t {
+                                    glyph_id: ctx.glyphs[i + 1].into(),
+                                    ..hb_glyph_info_t::default()
+                                };
+                                match_glyph(&mut info, value.get().to_u16())
```
dfrg, line 37: I guess we could just go `ctx.glyphs[i + 1] == value.get()` since this is just comparing glyph ids...

src/hb/ot/contextual.rs
```diff
@@ -152,7 +160,11 @@ impl WouldApply for ChainedSequenceContextFormat1<'_> {
                                 && rule.lookahead_glyph_count() == 0))
                             && ctx.glyphs.len() == input.len() + 1
                             && input.iter().enumerate().all(|(i, value)| {
-                                match_glyph(ctx.glyphs[i + 1], value.get().to_u16())
+                                let mut info = hb_glyph_info_t {
+                                    glyph_id: ctx.glyphs[i + 1].into(),
+                                    ..hb_glyph_info_t::default()
+                                };
+                                match_glyph(&mut info, value.get().to_u16())
```
dfrg, line 167: same here, just comparing gids

src/hb/ot_layout_gsubgpos.rs
```diff
@@ -432,7 +430,12 @@ where
     }
 
     #[inline]
-    pub fn match_(&self, info: &hb_glyph_info_t) -> match_t {
+    pub fn match_at(&mut self, idx: usize, from_out_info: bool) -> match_t {
```
dfrg, line 433: I like changing this to take an index but the bool is a bit confusing. I wonder if we can use an enum here (`MatchSource::Info` or `MatchSource::OutInfo`?) or just have two separate functions.

dfrg (approved):
Nice to see progress on this! LGTM with a couple suggestions.

src/hb/ot_layout_gsubgpos.rs
behdad: I don't like copy/pasted function. If you can figure out a way without duplication, that's my preference. Enum sounds good. I'll see about it.

src/hb/ot/contextual.rs
behdad: Humm. I like it this way (closer to HB code) if you don't mind.

src/hb/ot_layout_gsubgpos.rs
behdad: Fixed.

src/hb/ot/contextual.rs
behdad: I see the appeal though.

behdad: Thanks.

PR #125, opened 2025-08-09
[aat] general code cleanup

dfrg:
No functional changes, just making an effort toward cleaning up the code to make it easier to navigate and read. The focus here is on AAT since the code is smaller and I'm less likely to step on toes for other changes. If this looks good, I'll make changes incrementally to the rest of the codebase.

Does the following:
1. Moves all AAT files to the `aat` module and drops the aat_ prefix
2. Rustifies names, e.g. `feature_info_t` becomes `FeatureInfo`
3. We have modules, so drops prefixes from function names, e.g. `aat_layout::hb_aat_layout_position` becomes `aat::layout::position`
4. Similarly drops prefixes from constants: `HB_AAT_LAYOUT_FEATURE_TYPE_LETTER_CASE` -> `FEATURE_TYPE_LETTER_CASE` (these should probably move into a `feature_type` module)
5. Annotates many types/functions with original HB names, doc alias attributes and permalinks to relevant HB source code


behdad:
Thanks for the annotations.

PR #126, opened 2025-08-10
[Contextual] input-class cache in the `syllable()` buffer variable

behdad:
The cache facilities need to be hooked up to the LookupInfo still.


behdad:
The Context lookup part and the plumbing looks ready. The ChainContext is left TBD.

I see a nice speedup in NotoUrduNastaliq. BUT, this slows down EVERY sublookup application by adding a conditional `if cached`. I can measure that slowdown: Roboto benchmark gets slowed down from 21ms to 22.6ms. It's a pity to lose that much peformance.

behdad:
> this slows down EVERY sublookup application by adding a conditional `if cached`. I can measure that slowdown: Roboto benchmark gets slowed down from 21ms to 22.6ms. It's a pity to lose that much performance.

I tried a version with a single `match (self, cached)` instead of a conditional. It didn't improve things.

PR #131, opened 2025-08-11
[benches] Use criterion for benchmarks

dfrg:
Reworks the benchmarking setup.

* Drops the benches crate and adds benchmarks directly to harfrust
* Splits each text file into lines and shapes them independently
* Doesn't include the cold benchmarks. We can restore those later.


behdad:
Thanks. Numbers are generally consistent with HB benchmarks. The `hb` numbers are a bit higher than what we measure in HB, probably due to harfbuzz_rs overhead?

This is still much slower by default than the HB ones. But if it works for you, I'm happy.

LGTM. Thanks!

PR #133, opened 2025-08-11
Subtable dispatch through function pointers

dfrg:
Kinda ugly and not ready to merge yet, but worth doing some measurement.


behdad:
I measure some speedup as well. Yay.

PR #138, opened 2025-08-12
[GDEF] Use a bit-set for representing mark-filtering-sets

behdad:
HB does this.

Not cleaned up. Waiting for https://github.com/googlefonts/fontations/issues/1600

3% speedup in NotoNastaliqUrdu benchmark.

### DIFF
--- a/src/hb/ot/mod.rs
+++ b/src/hb/ot/mod.rs
@@ -1,6 +1,6 @@
 use super::buffer::GlyphPropsFlags;
 use super::ot_layout::TableIndex;
-use super::{common::TagExt, set_digest::hb_set_digest_t};
+use super::{common::TagExt};
 use crate::hb::hb_tag_t;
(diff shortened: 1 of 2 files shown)


behdad:
I see 1% slowdown in NotoNastalie; whereas in HB we don't see a slowdown. I still suggest we take it, since it improves NotoSansDuployan mega-font by over 10% in both HB & HR, even though we currently don't shape that font correctly.

First need to sort out no_std though.

behdad:
> I see 2% slowdown in NotoNastaliq; whereas in HB we don't see a slowdown.

This might be because the HB BitSet has more optimizations than Fontations U32Set.

behdad:
> 3% speedup in NotoNastaliqUrdu benchmark.

My benchmarks are a bit off today. I'll do better benchmarking another day.

PR #147, opened 2025-08-14
Micro-optimize buffer.push_str()

behdad:


src/hb/buffer.rs
```diff
@@ -1533,7 +1537,7 @@ impl hb_buffer_t {
         }
 
         for (i, c) in text.char_indices() {
-            self.add(c as u32, i as u32);
+            self.add_ensured(c as u32, i as u32);
```
rsheeter, line 1540: Nit: `add_ensured` doesn't convey much to me. It sounds slightly like it validates or checks something. In add it semi-clearly reads as "ensure there is enough space" based on context, here somewhat less so.

Spitballing, I wonder if we could implement Index/IndexMut and then write something akin to:

```rust
        for (i, c) in text.char_indices() {
            self[self.len] = hb_glyph_info_t::new(codepoint, cluster);
            self.len += 1;
        }
```

behdad: Thanks.https://github.com/harfbuzz/harfrust/pull/151

PR #152, opened 2025-08-14
Fix type inference issues with ClassDef::get()

dfrg:
And clean up all of the unnecessary conversions.


behdad:
Nice. I was wondering about these conversions.

dfrg:
A while back I changed the coverage tables to take `impl Into<GlyphId>` so they would accept any convertible type but the classdef getters still took `GlyphId16` so we needed a bunch of ugly checks.

PR #154, opened 2025-08-14
[ligature] Tune slow-path

behdad:
From HB.


behdad:
Oops. The fastpath is failing a couple tests. It doesn't happen in HB. Investigating.

PR #162, opened 2025-08-15
Micro-optimize advance widths

dfrg:
Retain the h_metrics array from hmtx so we don't parse/validate it for each advance width lookup.


behdad:
Nice.

I don't measure much difference, but good to have.

PR #164, opened 2025-08-15
Revert "Use the new fontations Value type for all GPOS tables (#161)"

dfrg:
This reverts commit 047f9a4b51d0f382f4d7e5b5b50dd0839be99271.


dfrg:
This seems to negatively affect perf in some measurements. Let's revert and revisit later.

PR #168, opened 2025-08-15
Cache props matching

behdad:
From https://github.com/harfbuzz/harfbuzz/pull/5420

I can't measure any speedups, not even with Duployan. Investigating.


behdad:
With latest additions, I measure no slowdown. Yay.

PR #169, opened 2025-08-15
Unify on new value record application

dfrg:
Updated treatment for pairpos1 and singlepos1/2 to match the work for pairpos2 in #167

Cleans up the code quite a bit so hopefully this one doesn't create any perf regressions.


dfrg:
I measured no perf change on this one.

behdad:
Looking good. Slightly faster even.

PR #177, opened 2025-08-16
Rustify hb_glyph_info_t and friends

dfrg:
This is a large patch but no functional changes.

* Renames `hb_glyph_info_t` -> `GlyphInfo` (for some reason, `hb_glyph_position_t` was already done?)
* Refactors all `_hb_glyph_info_` functions into methods on `GlyphInfo`
* Adds comments and doc alias attributes with the original HB names as well as permalinks to the relevant HB source for each function

An example of the change:

```rust
        if _hb_glyph_info_is_default_ignorable(info)
            && (self.ignore_zwnj || !_hb_glyph_info_is_zwnj(info))
            && (self.ignore_zwj || !_hb_glyph_info_is_zwj(info))
            && (self.ignore_hidden || !_hb_glyph_info_is_hidden(info))
```
becomes
```rust
        if info.is_default_ignorable()
            && (self.ignore_zwnj || !info.is_zwnj())
            && (self.ignore_zwj || !info.is_zwj())
            && (self.ignore_hidden || !info.is_hidden())
```


behdad:
Beautiful. Thanks.

I got a heart attack first thinking that maybe you moved all the methods to buffer.rs, but I see you did it exactly how I wanted to do. :)

dfrg:
> I got a heart attack first thinking that maybe you moved all the methods to buffer.rs, but I see you did it exactly how I wanted to do. :)

I figured you'd prefer if I kept everything in the same files and order. Luckily we have magical Rust impl blocks so we can add methods to types from anywhere :)

behdad:
> I figured you'd prefer if I kept everything in the same files and order. Luckily we have magical Rust impl blocks so we can add methods to types from anywhere :)

Thanks :). Yeah it's really cool this way. The reason I don't want them all on buffer is that most of those methods are only valid within certain var allocation times.

PR #178, opened 2025-08-16
Hide some methods on GlyphInfo

dfrg:
In #177, some of the functions originally had `pub` visibility and retained the same when I made them methods. These would be exposed in the public API which we don't want.. so make them `pub(crate)` to match the others.


behdad:
Thanks. There's a lot in the API that should become crate visible.

dfrg:
Agreed. We've been a bit liberal with pub.

PR #179, opened 2025-08-16
Branchfree set_digest

behdad:
Port from https://github.com/harfbuzz/harfbuzz/pull/5469


behdad:
Unlike in HB, this doesn't move the needle. Maybe clang was smarter than gcc in already doing this branchfree. I'll unroll the look and double-check.

PR #182, opened 2025-08-16
Don't panic! in the library

behdad:
Reverts previous commit's panics.

### DIFF
--- a/src/hb/ot/gpos/pair.rs
+++ b/src/hb/ot/gpos/pair.rs
@@ -23,7 +23,7 @@ impl Apply for PairPosFormat1<'_> {
                     &cache.coverage,
                 )?
             } else {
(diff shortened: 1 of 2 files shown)


src/hb/ot/gpos/pair.rs
```diff
@@ -23,7 +23,7 @@ impl Apply for PairPosFormat1<'_> {
                     &cache.coverage,
                 )?
             } else {
-                panic!(""); // Tell compiler this can't happen.
+                return None; // This can't actually happen.
```
LaurenzV, line 26: If you really want to indicate that it's not reachable, the canonical way of doing this would be to use `unreachable!()` (but this will also lead to a panic in case it is reached).

PR #184, opened 2025-08-17
Replace usage of char with u32

dfrg:
This is closer to HB. I thought we might see a small bump in perf from removing the fallible conversions but no joy. Still, I don't think font code should be policing the content of cmap so I suggest we merge this.


khaledhosny:
HarfBuzz also accepts invalid Unicode code points (outside of the valid 21 bit range) and applications use that as a trick to pass glyph IDs directly to HarfBuzz buffer (by implementing get_nominal_glyph() callback and intercepting the invalid code points, but I understand HarfRust does not currently have font callbacks).

PR #194, opened 2025-08-20
Add smaller, secondary subtable caches

dfrg:
This adds a secondary set of smaller "caches" for the PairPos and Ligature subtables. This is basically just storing small metadata from Coverage and ClassDef tables so we can avoid a bunch of reads and bounds checks. For ligature subtables, this keeps the "seconds" set digest in the smaller cache but that could be removed.

Based on #192


behdad:
Thanks. This speeds up GoogleSans by 20%.

PR #215, opened 2025-08-24
[aat] Towards adding bitset caches

behdad:
Depends on https://github.com/googlefonts/fontations/pull/1633


dfrg:
> With nostd, read-fonts doesn't have any U32Set. So we need to somehow disable the cache. It would be very intrusive to remove all mentions of U32Set though. So, maybe we can define it to a dummy type and make the cache always bypass.

Is it worth swapping in a set digest in the no_std case?

behdad:
> Is it worth swapping in a set digest in the no_std case?

Good question. I'll investigate.

dfrg:
I'll go over this in more detail when I consider how to merge the logic for kerx/kern. But generally LGTM now so let's land it.

PR #240, opened 2025-08-29
Try to micro-optimize AAT rearrangement

dfrg:
Does this help Indic perf?


dfrg:
Done but that was only part of the HR benchmarking code so shouldn't affect your measurements.

behdad:
Thanks. I still see slowdown, by about 1%.

behdad:
> Thanks for checking. LLVM probably does a good job of removing the bounds checks here in the original code.

Yeah. Versus the

PR #246, opened 2025-09-02
Minor optimizations for (Chain)ContextFormat2

dfrg:
We have some unused space in the external subtable cache so use it for these tables to avoid some fontations overhead. Seems to buy a little bit on Nastaliq according to HR benches.


behdad:
I measured about 2% improvement on Nastaliq.

PR #247, opened 2025-09-02
Always inline skippy next and match_at

dfrg:
These were popping up on the profile while shaping Roboto.


behdad:
Unfortunately in hb-harfrust this slows down Nataliq by 5%ish and shows no speedup in other cases.

Issue #256, opened 2025-09-10
Repeated pair positioning lookups could probably be optimized


valadaptive:
When doing shaping+layout for the version of Ubuntu Light used in egui ([Ubuntu-Light.ttf.zip](https://github.com/user-attachments/files/22253093/Ubuntu-Light.ttf.zip)), I noticed that we're spending a lot of time in `pair_pos2_cached` (20-30% of the shaping time!):

<img width="1035" height="853" alt="Image" src="https://github.com/user-attachments/assets/74a43a28-cfc4-4fde-b93a-a5b0a0b274fb" />

[Here's the samply profile.](https://github.com/user-attachments/files/22253171/profile.json.gz)

It seems that we re-parse the pair positioning table once per pair. Are we preparsing and caching these yet?

behdad:
Thanks.

I'm curious in your debugger on Linux. What is that?

valadaptive:
I should also mention how I run the profiler with the benchmarks. There might be a better way, but I look at the output from `cargo bench`, which provides the name of the binary:

```bash
[va_erie@Ampersand harfruzz]$ cargo bench
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.07s
     Running benches/shaping.rs (target/release/deps/shaping-8b2dd32552295ccf)
```

I can then run:

```bash
samply record target/release/deps/shaping-8b2dd32552295ccf --bench [BENCHMARK_NAME]
```

`--bench` (and all the arguments after the binary name) are passed directly to the binary; I think all binary artifacts produced by `cargo bench` need that argument passed (even if you don't pass a specific benchmark name).

dfrg:
This is the nature of OpenType layout running head first into fontations overhead. 

The general OTL algorithm is
```
foreach lookup:
    foreach glyph:
        foreach subtable:
```
So we pay the cost of subtable parsing over and over. We _can_ cache more with the trade off being additional memory use and more hand rolled parsing.

I’ll look into this.

valadaptive:
You can get full debug info in a release build by adding:

```toml
[profile.release]
debug = true
```

to your `Cargo.toml`. For a benchmark build (from the `cargo bench` command), it's:

```toml
[profile.bench]
debug = true
```

Cargo apparently also lets you [create custom profiles](https://stackoverflow.com/a/38804737), so you could create your own "release but with full debug info" profile.

behdad:
> You can get full debug info in a release build by adding:
> 
> [profile.release]
[...]

PR #260, opened 2025-09-10
Cache all fields from PairPos2 and avoid reading subtable

dfrg:
Attempt at #256

### DIFF
--- a/harfrust/src/hb/ot/gpos/mod.rs
+++ b/harfrust/src/hb/ot/gpos/mod.rs
@@ -11,6 +11,8 @@ mod mark;
 mod pair;
 mod single;
 
+pub(crate) use pair::apply_pair_pos2_direct;
+
 #[allow(unused_assignments)]
 fn apply_value(
     ctx: &mut hb_ot_apply_context_t,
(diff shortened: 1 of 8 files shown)


dfrg:
@valadaptive this avoids subtable reparsing entirely for pairpos2. Would you mind checking if this improves your profile?

valadaptive:
Yes! This speeds up shaping Ubuntu Light by ~8% in the benchmark.

behdad:
Do we still need this? Is it already done? Would the faster read-fonts obviate this?

dfrg:
I think this one _might_ still be useful but we’ll need to rebase and measure.

behdad:
This measures within noise for me in HB benchmarks. @valadaptive Any objections to closing?

valadaptive:
No objections to closing. Given the amount of perf work that's been done since I filed my original issue, it's probably worth benchmarking again to see if it's still worth optimizing this. I forgot what my benchmark setup was, unfortunately.

behdad:
> No objections to closing. Given the amount of perf work that's been done since I filed my original issue, it's probably worth benchmarking again to see if it's still worth optimizing this. I forgot what my benchmark setup was, unfortunately.

Thanks. I tested with Ubuntu-Regular, and this patch now slows things down by 3%. Let's close.

Issue #91, opened 2025-07-24
Export for GlyphExtents and Unicode Buffer internals


asibahi:
Hello

I am trying to redo the wasm shaper for HarfRust but as an external crate on top of it and `skrifa`. I am still on the phase of trying to port the code and I have not kept up with the changes from rustybuzz to HarfRust.

A few roadblocks are the following:

1. Getting the GlyphExtents data: One of the functions require Glyph Extents. This does not seem to be exposed. It *might* be just as simple as adding it to a `pub use` to `lib.rs`.

2. Getting the internals of `UnicodeBuffer`.  This needs to be reset this from raw data coming from the module. IF there is a better way to do this I am all ears.

```
store_data.buffer.len = buffer.length as usize;

store_data.buffer.info.clear();
store_data.buffer.info.extend_from_slice(bytemuck::cast_slice(
	&mem_data[buffer.info as usize..buffer.info as usize + array_length],
));

store_data.buffer.pos.clear();
store_data.buffer.pos.extend_from_slice(bytemuck::cast_slice(
	&mem_data[buffer.position as usize..buffer.position as usize + array_length],
));
```

3. Last answer least: `hb_font_t`, exported as `Shaper`, seems to have lost the inquiries about specific glyphs and tables. I am not quite sure how to access these. (I have not thought about this too much, and it might be as simple as just passing the `FontRef` from `read_fonts` separately.

Am I missing something obvious in regards to these points or are they something the crate can export?

behdad:
Hello hello.

Before you spend much time on this, I like to update you on my current plans for the wasm shaper. Basically, a full rehaul.

1. Get rid of the font_t abstraction. Everything will be in font units, like the rest of the font.
2. Rename the API to use a prefix. Not sure what. Following up in https://github.com/harfbuzz/harfbuzz/issues/4457

dfrg:
Skrifa has the [GlyphMetrics::bounds](https://docs.rs/skrifa/latest/skrifa/metrics/struct.GlyphMetrics.html#method.bounds) method which is currently more accurate than the internal extents code we use in HR since it supports CFF fonts and variations.

PR #298, opened 2025-11-12
Use a single Buffer type

dfrg:
This removes `GlyphBuffer` and `UnicodeBuffer`, replacing them with a single `Buffer` type to more closely match HB.

Pros:
* Closer to HB API
* Don't need to use `Option<UnicodeBuffer>` and `take()` to reuse buffers
* Possible to make an HB compatible FFI without copies
* Enable support for HB style tracing messages (#280)
* Toward support for WASM shaping in a separate crate (#91)
* Trivial to build the GlyphBuffer/UnicodeBuffer API on top

Cons:
* Larger number of methods so proper usage is less discoverable
* More methods are now fallible since they depend on the current content type

TODOs:
* Probably missed some content type checks but I think I found the important ones
* Replace asserts with errors... deferring since this PR is already large enough

### DIFF
--- a/src/hb/buffer.rs
+++ b/src/hb/buffer.rs
@@ -6,8 +6,8 @@ use read_fonts::types::{GlyphId, GlyphId16};
 
 use super::buffer::glyph_flag::{SAFE_TO_INSERT_TATWEEL, UNSAFE_TO_BREAK, UNSAFE_TO_CONCAT};
 use super::face::hb_glyph_extents_t;
+use super::hb_mask_t;
 use super::unicode::CharExt;
-use super::{hb_font_t, hb_mask_t};
 use crate::hb::set_digest::hb_set_digest_t;
 use crate::hb::unicode::Codepoint;
 use crate::{script, BufferClusterLevel, BufferFlags, Direction, Language, Script, SerializeFlags};
@@ -389,6 +389,17 @@ impl GlyphInfo {
     }
 }
 
+/// The type of [`Buffer`] contents.
+///
+/// HarfBuzz calls this `hb_buffer_content_type_t`.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum BufferContentType {
+    /// The buffer contains input characters (before shaping).
+    Unicode,
+    /// The buffer contains output glyphs (after shaping).
+    Glyphs,
+}
+
 pub type hb_buffer_cluster_level_t = u32;
 pub const HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES: u32 = 0;
 pub const HB_BUFFER_CLUSTER_LEVEL_MONOTONE_CHARACTERS: u32 = 1;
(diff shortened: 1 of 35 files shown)


src/hb/buffer.rs
```diff
@@ -389,80 +389,85 @@ impl GlyphInfo {
     }
 }
 
+/// The type of [`Buffer`] contents.
+///
+/// HarfBuzz calls this `hb_buffer_content_type_t`.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum BufferContentType {
```
behdad: HarfBuzz also has a content type Invalid. That's used when you reset the buffer. Lemme see how you handle that...

src/hb/buffer.rs
```diff
@@ -484,8 +489,368 @@ impl hb_buffer_t {
...
+            };
+            self.len += 1;
+        }
+    }
+
+    /// Appends a character to a buffer with the given cluster value.
+    #[inline]
+    pub fn push(&mut self, codepoint: char, cluster: u32) {
+        if self.content_type == Some(BufferContentType::Glyphs) {
+            // TODO: return an error
```
behdad: This function should be renamed to `push_char` or something. Otherwise it's hard to differentiate it from `add`.

behdad (approved):
LGTM with one comment inline. I'll comment separately on the assert vs error.

behdad:
> * Probably missed some content type checks but I think I found the important ones

Actually now I see a missing check in `guess_segment_properties` as well. HB has `assert_unicode` there.

> * Replace asserts with errors... deferring since this PR is already large enough

I'm a proponent of assert / panic as these are preconditions. Errors that can never happen in correctly-constructed code end up just polluting the code IMO.

rsheeter:
> Errors that can never happen in correctly-constructed code end up just polluting the code IMO.

Making apis _look_ infallible when they are fallible is hostile and violates the expectation that if your Rust code compiles it's likely to work.

Ideally we would structure the types so the error can't happen - nice if possible - so we don't need assert. If not then the failure is a valid outcome and Error seems correct, the user gets to decide what to do if they held it wrong and is made explicitly aware that it's possible to hold it wrong.

behdad:
> Making apis _look_ infallible when they are fallible is hostile and violates the expectation that if your Rust code compiles it's likely to work.

So.. you don't believe in invariants, then?

src/hb/buffer.rs
dfrg: Yeah, this is simply more idiomatic when dealing with enums that have some sort of invalid state. We should probably do the same for the others.

src/hb/buffer.rs
behdad: We want to keep `add()`, for lowlevel buffer work.

behdad:
We should resurrect and land this, to unblock https://github.com/harfbuzz/harfrust/pull/280

simoncozens:
Yes please...

PR #302, opened 2025-11-28
Change logo to PNG with transparent background

behdad:
It's huge. Not useful.

cc @khaledhosny


behdad:
It does look better though.

PR #306, opened 2025-12-08
Avoid panic when saving syllable indices

### LINKED ISSUE #305: Panic in src/hb/ot_shaper_indic.rs when shaping text with Noto Sans Kannada
This was reported here: https://github.com/pop-os/cosmic-epoch/issues/2597

To reproduce the panic, download [song.txt](https://github.com/user-attachments/files/24040229/song.txt) and run the following command:

```
cargo run --example shape /usr/share/fonts/truetype/noto/NotoSansKannada-Regular.ttf "$(cat song.txt)"
```

You should get similar output to this:

```
thread 'main' (1448927) panicked at src/hb/ot_shaper_indic.rs:1106:65:
called `Result::unwrap()` on an `Err` value: TryFromIntError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

dfrg:
We use a single byte to temporarily save the indices of the pre-sorted characters in each syllable. This overflows and causes a panic if the length of the syllable exceeds 255. However, we don't actually read these values back if the length exceeds 127 (see https://github.com/harfbuzz/harfrust/blob/904ed514cd50f2d345745bc9f6346c54630171ce/src/hb/ot_shaper_indic.rs#L1179) so the overflow is irrelevant and this patch removes the checked operation.

Fixes #305

### DIFF
--- a/src/hb/ot_shaper_indic.rs
+++ b/src/hb/ot_shaper_indic.rs
@@ -1,6 +1,5 @@
 use alloc::boxed::Box;
 use core::cmp;
-use core::convert::TryFrom;
 use core::ops::Range;
 
 use read_fonts::types::GlyphId;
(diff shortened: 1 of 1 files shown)


dfrg:
https://crates.io/crates/harfrust/0.4.1

Issue #308, opened 2025-12-09
[Feature Request] Use a generic TableProvider instead of directly depending on FontRef


2A5F:
This will allow the use of other data sources, such as dwrite.

LGFae:
I was looking into implementing a renderer where we only keep the information necessary for shaping in memory, and read and discard the glyph outline/bitmap information (since we cache it after the first time). This would really improve memory efficiency in my applications.

But that's only possible if I can use my own implementation of a `TableProvider`, instead of `FontRef`.

PR #309, opened 2025-12-10
Replace FontRef with TableProvider

2A5F:
This is a breaking change, so for now it’s a draft, mainly to verify feasibility.

#308


behdad:
@dfrg What's your stance on this now with new Font API and all?

PR #334, opened 2026-02-15
Bug fixes discovered using AI against HB

behdad:
The decompose() and decompose_current_character() functions had inverted conditions compared to HarfBuzz. When shortest mode is enabled and a glyph exists, should output immediately without recursive decomposition.

This fixes incorrect normalization behavior across multiple scripts.

This bug was discovered and fixed by Claude Code.

### DIFF
--- a/src/hb/buffer.rs
+++ b/src/hb/buffer.rs
@@ -1005,7 +1005,7 @@ impl hb_buffer_t {
 
         // Extend start
         if cluster != self.info[start].cluster {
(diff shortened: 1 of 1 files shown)


khaledhosny:
Needs a test, I think.

behdad:
I believe so. I'll see what I can do.

behdad:
It found a copy/paste error as well. Again, hard to test, but definitely a bug.

PR #336, opened 2026-02-28
sumplify _hb_ot_shape_normalize a little

asibahi:
- get rid of some double pointers
- more idiomatic rust re-overridden decompose and compose

I was going through this and I was wondering about the double pointers, which I assume was to satisfy the borrow checker. `cargo test` passes all tests after this change, so no functional changes as far as I can tell.


behdad:
Thanks. I'll review and if necessary rebase after we decide on https://github.com/harfbuzz/harfrust/pull/332#issuecomment-3979212181

PR #353, opened 2026-04-24
[fuzz] turn off strip/lto

cmyr:
This was ~possibly~ probably causing us to fail to build on oss-fuzz?

~(edit: I'm now not convinced this is the issue since fontations doesn't have to do this? needs more investigation)~


behdad:
What's the failure?

cmyr:
https://github.com/google/oss-fuzz/actions/runs/24145916341/job/72892193106?pr=15284

cmyr:
Okay I investigated this further, and I think this is the problem; the issue is that these options are explicitly enabled  in the workspace Cargo.toml, so we need to override that in the fuzz targets.

PR #355, opened 2026-05-19
[POC] override batch advance widths

dfrg:
Figuring out how to do this.

### DIFF
--- a/harfrust/src/lib.rs
+++ b/harfrust/src/lib.rs
@@ -23,7 +23,7 @@ pub use read_fonts::{types::Tag, FontRef};
 
 pub use hb::buffer::{GlyphBuffer, GlyphInfo, GlyphPosition, UnicodeBuffer};
 pub use hb::common::{script, Direction, Feature, Language, Script, Variation};
-pub use hb::face::{hb_font_t as Shaper, ShaperBuilder, ShaperData, ShaperInstance};
+pub use hb::face::{hb_font_t as Shaper, ShaperBuilder, ShaperData, ShaperInstance, FontOverrides};
 pub use hb::ot_shape_plan::{hb_ot_shape_plan_t as ShapePlan, ShapePlanKey};
 
 /// Type alias for a normalized variation coordinate.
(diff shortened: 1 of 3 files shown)


harfrust/src/lib.rs
```diff
@@ -117,3 +117,41 @@ bitflags::bitflags! {
...
+            advance_stride: c_uint,
+            user_data: *mut (),
+        ) {
+            //
+        }
+
+        struct Dummy;
+
+        impl FontOverrides for Dummy {
+            fn advance_widths<'a>(&mut self, widths: impl Iterator<Item = (u32, &'a mut i32)>) {
```
dfrg, line 152: This is the ideal Rust style interface but we need the internal pointers to the buffer to map this to HB font funcs?

behdad: Maybe the Chrome ffi layer can do it unsafely? No? @drott

drott: There is no strict safety requirement on the FFI we write, that's correct. How would that help in rephrasing here?

behdad: Humm. You're right that it doesn't help much.

So, if we add an iterator-based API to Skia, then translate the iterator, that would work, but probably cumbersome to translate an iterator to C++ one?

behdad: Or as Chad noted in chat, allocating memory would help.

drott: How about something in between? I am not sure how far HarfBuzz / HarfRust collect glyphs and whether this absolutely needs to scale with text size, but how about a fixed size batch buffer somewhere, without re-allocating, but enough space for a _good amount_ of character->glyph mapping, where _good amount_ means sizes we see in the wild?

How is the batch collected? A full scan over the incoming shape-task buffer?

dfrg: I went with a different approach where the callback receives an object that can be converted into either a safe rust iterator or a struct containing the pointers and strides as HB expects. The hb-hr layer can choose the latter and invoke the function provided by the client directly.

dfrg:
New (functional) approach at #365

PR #358, opened 2026-05-21
Let read-fonts decide whether libm is needed

drott:
Read-fonts only needs libm if std is not available. We do not want to hard-required libm for read-fonts here, compare [1].

[1] https://github.com/googlefonts/fontations/blob/main/read-fonts/Cargo.toml#L33C1-L33C26

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ inherits = "release"
 
 [workspace.dependencies]
 harfrust = { version = "0.6.1", path = "./harfrust", default-features = false }
-read-fonts = { version = "0.39.0", default-features = false, features = ["libm"] }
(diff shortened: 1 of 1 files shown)


behdad:
lgtm but I'm no expert on these matters.

dfrg:
Note that the libm feature optionally links the core_math crate which is currently an unconditional dependency for harfrust. Also, prior to this change, `cargo build --no-default-features` worked for harfrust but now fails so we're potentially breaking some users.

We should copy the read-fonts setup, adding a libm feature to harfrust, passing that along to read-fonts.

drott:
Right, ugh, we missed this link. Thanks.

dfrg:
Fix in #362

PR #361, opened 2026-05-21
Refactor ot shape context functions into methods

dfrg:
No functional changes. Simply adding methods to lay the groundwork for font funcs and unicode funcs since we'll need to add some type parameters to this code and it's much simpler if we can tack them onto an impl block.

As a bonus, adds permalinks to associated HB sources :)


cmyr (approved):
looks good!

behdad (approved):
Great. Thanks.

PR #362, opened 2026-05-21
Make core_maths an optional dependency

dfrg:
through the libm feature and forward the same to read-fonts

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,3 +37,20 @@ jobs:
 
     - name: Clippy
       run: cargo clippy --all-features --all-targets -- -D warnings
+
+  check-no-std:
+    name: cargo check no std
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: install stable toolchain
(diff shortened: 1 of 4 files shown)


dfrg:
Looking again, I'm not even sure if we use any libm functions in this crate...

dfrg:
> Looking again, I'm not even sure if we use any libm functions in this crate...

trak table uses f32::round(). We could just polyfill this but we need core_maths for no_std read-fonts anyway so I don't think it's worth it.

drott:
LGTM, thanks for fixing this. I need to test whether this still works with Chromium's crate importer. I assume it would, as the math functions would come from `std` in any case, and we can keep the feature off.

dfrg:
> LGTM, thanks for fixing this. I need to test whether this still works with Chromium's crate importer. I assume it would, as the math functions would come from std in any case, and we can keep the feature off.

Right. If you just keep the default features enabled (or enable std) then everything should just work.

We should do a 0.7.0 release and yank 0.6.2 whenever you no longer need it since these feature adjustments are technically breaking changes and we'll probably get some complaints rolling in at some point :)

PR #364, opened 2026-05-21
New ShapeOptions type

dfrg:
We need some mechanism to specify font funcs, unicode funcs and scale when shaping.

This introduces a new `ShapeOptions` type which is passed to the `shape` method. The type acts as a builder, currently accepting shape plan, point size and features. Eventually this will include scale, font funcs and unicode funcs (and maybe others?).

The `shape_with_plan` method goes away since it is now redundant. `ShaperBuilder::point_size` moves to `ShapeOptions`.

Based on #361 because both sets of changes touch the shaping context.


cmyr (approved):
looks good!

PR #365, opened 2026-05-25
Add support for font funcs

dfrg:
Add a new FontFuncs trait, support for configuring it in ShapeOptions and all the lovely plumbing to make use of it.

The patch is larger than I would like but the changes are fairly invasive because the callbacks need to be passed (mutably!) through a lot of layers which required some reorganizing of code and some additional lifetimes.

### DIFF
--- a/harfrust/src/hb/face.rs
+++ b/harfrust/src/hb/face.rs
@@ -4,6 +4,7 @@ use smallvec::SmallVec;
 
 use super::aat::AatTables;
 use super::charmap::{cache_t as cmap_cache_t, Charmap};
+use super::font_funcs::{DummyFontFuncs, FontFuncsDispatch};
 use super::glyph_metrics::GlyphMetrics;
 use super::glyph_names::GlyphNames;
 use super::ot::{LayoutTable, OtCache, OtTables};
@@ -13,6 +14,8 @@ use crate::hb::aat::AatCache;
 use crate::hb::buffer::hb_buffer_t;
 use crate::hb::tables::TableRanges;
 use crate::{script, Feature, GlyphBuffer, NormalizedCoord, ShapePlan, UnicodeBuffer, Variation};
+
+pub use super::font_funcs::{AdvanceWidthBatch, BuiltinFontFuncs, FontFuncs, RawAdvanceWidthBatch};
 
 /// Data required for shaping with a single font.
 pub struct ShaperData {
@@ -196,7 +199,7 @@ impl<'a> ShaperBuilder<'a> {
     pub fn build(self) -> crate::Shaper<'a> {
         let font = self.font;
         let units_per_em = self.data.table_ranges.units_per_em;
-        let charmap = Charmap::new(&font, &self.data.table_ranges, &self.data.cmap_cache);
+        let charmap = Charmap::new(&font, &self.data.table_ranges);
         let glyph_metrics = GlyphMetrics::new(&font, &self.data.table_ranges);
         let (coords, feature_variations) = self
             .instance
@@ -214,6 +217,7 @@ impl<'a> ShaperBuilder<'a> {
             font,
             units_per_em,
             charmap,
+            cmap_cache: &self.data.cmap_cache,
             glyph_metrics,
             ot_tables,
             aat_tables,
@@ -227,6 +231,7 @@ pub struct ShapeOptions<'a> {
     plan: Option<&'a ShapePlan>,
     point_size: Option<f32>,
     features: &'a [Feature],
+    font_funcs: Option<&'a mut (dyn FontFuncs + 'a)>,
 }
 
 impl<'a> ShapeOptions<'a> {
@@ -255,6 +260,12 @@ impl<'a> ShapeOptions<'a> {
         self.features = features;
         self
     }
+
+    /// Sets optional font functions used for shaping.
+    pub fn font_funcs(mut self, funcs: Option<&'a mut (dyn FontFuncs + 'a)>) -> Self {
+        self.font_funcs = funcs;
+        self
+    }
 }
 
 /// A configured shaper.
@@ -263,6 +274,7 @@ pub struct hb_font_t<'a> {
     pub(crate) font: FontRef<'a>,
     pub(crate) units_per_em: u16,
     charmap: Charmap<'a>,
+    pub(crate) cmap_cache: &'a cmap_cache_t,
     glyph_metrics: GlyphMetrics<'a>,
     pub(crate) ot_tables: OtTables<'a>,
     pub(crate) aat_tables: AatTables<'a>,
@@ -284,7 +296,15 @@ impl<'a> crate::Shaper<'a> {
     ///
     /// Consumes the buffer. You can then run [`GlyphBuffer::clear`] to get the [`UnicodeBuffer`] back
     /// without allocating a new one.
-    pub fn shape(&self, buffer: UnicodeBuffer, options: ShapeOptions) -> GlyphBuffer {
+    ///
+    /// If a plan is provided, it is up to the caller to ensure that the shape plan matches the
+    /// properties of the provided buffer, otherwise the shaping result will likely be incorrect.
+    ///
+    /// # Panics
+    ///
+    /// Will panic when debugging assertions are enabled if the buffer and plan have mismatched
+    /// properties.    
+    pub fn shape(&self, buffer: UnicodeBuffer, options: ShapeOptions<'_>) -> GlyphBuffer {
         if let Some(plan) = options.plan {
             self.shape_with_plan(plan, buffer, options)
         } else {
@@ -299,23 +319,11 @@ impl<'a> crate::Shaper<'a> {
         }
     }
 
-    /// Shapes the buffer content using the provided font and plan.
-    ///
-    /// Consumes the buffer. You can then run [`GlyphBuffer::clear`] to get the [`UnicodeBuffer`] back
-    /// without allocating a new one.
-    ///
-    /// It is up to the caller to ensure that the shape plan matches the properties of the provided
-    /// buffer, otherwise the shaping result will likely be incorrect.
-    ///
-    /// # Panics
-    ///
-    /// Will panic when debugging assertions are enabled if the buffer and plan have mismatched
-    /// properties.
     fn shape_with_plan(
         &self,
         plan: &ShapePlan,
         buffer: UnicodeBuffer,
-        options: ShapeOptions,
+        options: ShapeOptions<'_>,
     ) -> GlyphBuffer {
         let mut buffer = buffer.0;
         buffer.enter();
@@ -336,6 +344,12 @@ impl<'a> crate::Shaper<'a> {
         if buffer.len > 0 {
             // Save the original direction, we use it later.
             let target_direction = buffer.direction;
+            let (font_funcs, has_custom_funcs): (&mut (dyn FontFuncs + '_), bool) =
+                match options.font_funcs {
+                    Some(user_font_funcs) => (user_font_funcs, true),
+                    None => (&mut DummyFontFuncs, false),
+                };
+            let mut font_funcs = FontFuncsDispatch::new(self, font_funcs, has_custom_funcs);
             OtShapeContext {
                 plan,
                 face: self,
@@ -343,6 +357,7 @@ impl<'a> crate::Shaper<'a> {
                 target_direction,
                 features: options.features,
                 point_size: options.point_size,
+                font_funcs: &mut font_funcs,
             }
             .shape_internal();
         }
@@ -350,10 +365,6 @@ impl<'a> crate::Shaper<'a> {
         buffer.leave();
 
         GlyphBuffer(buffer)
-    }
-
-    pub(crate) fn has_glyph(&self, c: u32) -> bool {
-        self.get_nominal_glyph(c).is_some()
     }
 
     pub(crate) fn get_nominal_glyph(&self, c: u32) -> Option<GlyphId> {
@@ -369,6 +380,7 @@ impl<'a> crate::Shaper<'a> {
             .advance_width(glyph, self.ot_tables.coords)
             .unwrap_or_default()
     }
+
     pub(crate) fn glyph_h_advances(&self, buffer: &mut hb_buffer_t) {
         self.glyph_metrics
             .populate_advance_widths(buffer, self.ot_tables.coords);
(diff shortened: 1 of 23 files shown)


harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+use super::face::{hb_font_t, GlyphExtents};
+
+/// Raw C-style view over a batch of glyph ids and advance widths.
+#[derive(Clone, Copy, Debug)]
+pub struct RawAdvanceWidthBatch {
+    /// Number of batch entries.
+    pub len: usize,
+    /// Pointer to glyph ids (read-only).
+    pub gids: *const u32,
+    /// Pointer to horizontal advances (writable).
```
drott, line 17: Suggest to clarify unit/scaling expectations of the value? I assume, font units, or is this like `hb_position_t`, 16.16 fixed point? 
Also, perhaps add a comment on that it's up to the explanation to round for hinting?

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+        builtin.vertical_origin(glyph)
+    }
+
+    /// Glyph extents callback.
+    fn extents(&mut self, builtin: &BuiltinFontFuncs, glyph: GlyphId) -> Option<GlyphExtents> {
+        builtin.extents(glyph)
+    }
+}
+
+pub struct DummyFontFuncs;
```
drott:
```suggestion
pub struct EmptyFontFuncs;
```

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+
+impl FontFuncs for DummyFontFuncs {}
```
drott:
```suggestion
impl FontFuncs for EmptyFontFuncs {}
```

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+    builtin: BuiltinFontFuncs<'a>,
+    funcs: &'u mut (dyn FontFuncs + 'u),
+    has_custom_funcs: bool,
+}
+
+impl<'a, 'u> FontFuncsDispatch<'a, 'u> {
+    pub(crate) fn new(
+        face: &'a hb_font_t<'a>,
+        funcs: &'u mut (dyn FontFuncs + 'u),
+        has_custom_funcs: bool,
```
drott: Does the `has_custom_funcs` mean that the `hb_font_t` has custom fonts, or that `BuiltinFontFuncs` are used vs. the ones from `hb_font_t`? I am slightly confused, because there is also a situation, in which the "built-in" font-funcs of the `hb_font_t` are used, i.e. HarfBuzz' OT font funcs. Perhaps clarify the bool by changing name or adding comment?

Second thought, depending on the above, is the bool needed, or can it be expressed through an `Option`?

harfrust/src/hb/ot_shape.rs
```diff
@@ -313,17 +313,32 @@ impl<'a> hb_ot_shape_planner_t<'a> {
...
+        if self.buffer.len == 0 {
+            return;
+        }
+
+        if !self.font_funcs.has_custom_funcs() {
+            self.face.glyph_h_advances(self.buffer);
+            return;
+        }
+
+        let batch = AdvanceWidthBatch::new(self.buffer);
```
drott:
```suggestion
        let batched_advances = AdvanceWidthBatch::new(self.buffer);
```
Maybe slightly clearer name?

harfrust/src/hb/font_funcs.rs
drott: After reading `OtShapeContext` - looks like this boolean means whether `funcs` is set to something. Not sure if possible, but could this be 

```
pub(crate) struct FontFuncsDispatch<'a, 'u> {
    face: &'a hb_font_t<'a>,
    funcs: Option<&'u mut (dyn FontFuncs + 'u)>,
}
```
then, would that make sense?

harfrust/src/hb/face.rs
```diff
@@ -336,13 +344,20 @@ impl<'a> crate::Shaper<'a> {
         if buffer.len > 0 {
             // Save the original direction, we use it later.
             let target_direction = buffer.direction;
+            let (font_funcs, has_custom_funcs): (&mut (dyn FontFuncs + '_), bool) =
```
drott: This seems complicated vs. having the `Option< &mut dyn FontFuncs>` on `FontFuncsDispatch`, but I might be missing something.

harfrust/src/hb/ot_shaper_arabic.rs
```diff
@@ -452,7 +468,7 @@ fn record_stch(plan: &hb_ot_shape_plan_t, _: &hb_font_t, buffer: &mut hb_buffer_
     false
 }
 
-fn apply_stch(face: &hb_font_t, buffer: &mut hb_buffer_t) {
+fn apply_stch(face: &mut FontFuncsDispatch, buffer: &mut hb_buffer_t) {
```
drott, line 471: Here and in other places `face: ` for the font functions argument does not seem 100% matching anymore. `font_funcs` maybe?

harfrust/src/hb/ot_shaper_indic.rs
```diff
@@ -637,7 +641,7 @@ fn setup_syllables(_: &hb_ot_shape_plan_t, _: &hb_font_t, buffer: &mut hb_buffer
 
 fn initial_reordering(
     plan: &hb_ot_shape_plan_t,
-    face: &hb_font_t,
+    font_funcs: &mut FontFuncsDispatch,
```
drott, line 644: Here you did rename it to `font_funcs`, suggest to do that everywhere where `face` is now `FontFuncsDispatch`.

harfrust/tests/font_funcs.rs
```diff
@@ -0,0 +1,397 @@
...
+    let (baseline, with_funcs) = with_test_shaper(|shaper| {
+        let baseline = shaper.shape(buffer_with_text("abc"), ShapeOptions::new());
+        let with_funcs = shaper.shape(
+            buffer_with_text("abc"),
+            ShapeOptions::new().font_funcs(Some(&mut funcs)),
+        );
+        (baseline, with_funcs)
+    });
+
+    assert!(funcs.nominal_calls > 0);
```
drott, line 92:
```suggestion
    assert!(funcs.nominal_calls >= 2);
```
We do have 3 glyphs that we look up, perhaps we can raise the test assertion here?

harfrust/tests/font_funcs.rs
```diff
@@ -0,0 +1,397 @@
...
+    let mut funcs = BatchAdvanceFuncs { batch_calls: 0 };
+
+    let glyphs = with_test_shaper(|shaper| {
+        shaper.shape(
+            buffer_with_text("abc"),
+            ShapeOptions::new().font_funcs(Some(&mut funcs)),
+        )
+    });
+
+    assert!(funcs.batch_calls > 0);
```
drott, line 132: Ditto, `>= 2`.

harfrust/tests/font_funcs.rs
```diff
@@ -0,0 +1,397 @@
...
+    assert!(!glyphs.glyph_positions().is_empty());
+    assert!(glyphs
+        .glyph_positions()
+        .iter()
+        .all(|pos| pos.x_advance == 777));
+}
+
+#[test]
+fn font_funcs_batch_advance_uses_individual_override_by_default() {
```
drott: Don't fully understand this test - is the idea that for the other non `advance_width` functions, the built-ins are called? How is this tested?

harfrust/tests/font_funcs.rs
```diff
@@ -0,0 +1,397 @@
...
+    buffer.push_str("abc");
+    buffer.set_direction(Direction::TopToBottom);
+    buffer.guess_segment_properties();
+    buffer.set_direction(Direction::TopToBottom);
+
+    let _ = with_test_shaper(|shaper| {
+        shaper.shape(buffer, ShapeOptions::new().font_funcs(Some(&mut funcs)))
+    });
+
+    assert!(funcs.v_origin_calls > 0);
```
drott: `>=2`

harfrust/tests/font_funcs.rs
```diff
@@ -0,0 +1,397 @@
...
+    let mut buffer = UnicodeBuffer::new();
+    buffer.push_str("abc");
+    buffer.set_direction(Direction::TopToBottom);
+    buffer.guess_segment_properties();
+
+    let glyphs = with_test_shaper(|shaper| {
+        shaper.shape(buffer, ShapeOptions::new().font_funcs(Some(&mut funcs)))
+    });
+
+    assert!(funcs.advance_height_calls > 0);
```
drott: Raise number above 0?

drott (commented):
Thanks for doing this, looks very suitable to me - I'll see if I can plumb this through in HarfBuzz or sync with Behdad and see if we can connect it to Blink.

harfrust/src/hb/font_funcs.rs
dfrg: Will clarify. The intention is to match HB behavior in that the values provided should match the requested scale factor(s). Notably, scaling is not included in this PR but I plan to do that next.

harfrust/src/hb/font_funcs.rs
dfrg: Yes, that would be slightly cleaner but would add a branch to every callback which is what I was explicitly trying to avoid since we’re fighting for cycles at this point :)

This bool is only used at one call site to enable an internal fast path that proved to be quite valuable in profiling so I wanted to preserve it.

harfrust/src/hb/face.rs
dfrg: The same reasoning as above wrt branching on callbacks. Some of these are in hot code so IMO avoiding an additional branch justifies the complexity.

harfrust/src/hb/ot_shaper_arabic.rs
dfrg: Agree completely. This patch is already large so I tried to avoid renaming lots of things to keep the diff smaller and focused on functional changes. I intend to do a pure naming/structural pass as a follow up.

harfrust/tests/font_funcs.rs
dfrg: The intention is to ensure that if you override `advance_width` but not `populate_advance_widths` that the latter calls the former. At one point, the logic wasn’t working that way so I added this test to catch regressions.

dfrg:
There is likely a cleaner overall design possible here once the read-fonts Font API changes (googlefonts/fontations#1852) land but I didn’t want to block on that.

harfrust/src/hb/font_funcs.rs
drott: Not sure we're really at the cycles level yet - I think allocations are still more of a problem. I believe the branching here is  rather constant in which branch it takes, i.e. should behave predictably after the CPU encounters it the first time? I would lean towards readability here.

harfrust/src/hb/ot_shaper_arabic.rs
drott: Ack.

harfrust/tests/font_funcs.rs
drott: Ok, I get it now. Maybe `s/individual/single_glyph/` in the test name?

harfrust/src/hb/font_funcs.rs
behdad: > This bool is only used at one call site to enable an internal fast path that proved to be quite valuable in profiling so I wanted to preserve it.

In HB we track whether each callback is overridden and can fastpath based on that. I suppose it's not possible here.

Which fastpath did you find this useful for?

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+        self.face.glyph_h_advance(glyph)
+    }
+
+    /// Returns the vertical advance for a glyph.
+    pub fn advance_height(&self, glyph: GlyphId) -> i32 {
+        self.face.glyph_v_advance(glyph)
+    }
+
+    /// Returns the horizontal origin for a glyph.
+    pub fn horizontal_origin(&self, glyph: GlyphId) -> i32 {
```
behdad: This is a misnomer in HR. In HB, `h_origin` means something else. What this value really is, is `v_origin_x`...

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+        self.advance_width(glyph) / 2
+    }
+
+    /// Returns the vertical origin for a glyph.
+    pub fn vertical_origin(&self, glyph: GlyphId) -> i32 {
```
behdad: And `v_origin_y`. Just documenting for now. I know this is pre-existing naming, and fine with me to keep for now and clean up later.

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+        for (glyph, advance) in batch {
+            *advance = self.face.glyph_h_advance(glyph);
+        }
+    }
+}
+
+/// Customizable font callback surface.
+pub trait FontFuncs {
+    /// Nominal character-to-glyph mapping callback.
+    fn nominal_glyph(&mut self, builtin: &BuiltinFontFuncs, c: u32) -> Option<GlyphId> {
```
behdad, line 167: I'm not sure I understand the role of `builtin` here exactly. Can it be changed to take a `parent` instead, ala HB? Such that one can override one func only if desired?

harfrust/src/hb/font_funcs.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    #[inline(always)]
+    pub(crate) fn advance_height(&mut self, glyph: GlyphId) -> i32 {
+        self.funcs.advance_height(&self.builtin, glyph)
+    }
+
+    #[inline(always)]
+    pub(crate) fn horizontal_origin(&mut self, glyph: GlyphId) -> i32 {
+        self.advance_width(glyph) / 2
```
behdad: Why does this not delegate?

behdad (approved):
Looks good to me. Left a few questions. Thanks.

harfrust/src/hb/font_funcs.rs
dfrg: oof, yes. I'm going to drop this and change `vertical_origin` to return both coordinates to match HB/Chrome.

behdad:
Okay, hb-hr POC in https://github.com/harfbuzz/harfbuzz/pull/6002

harfrust/src/hb/font_funcs.rs
dfrg: > Not sure we're really at the cycles level yet - I think allocations are still more of a problem. I believe the branching here is rather constant in which branch it takes, i.e. should behave predictably after the CPU encounters it the first time? I would lean towards readability here.

Easy enough to change later so I'll defer to the readability argument.

> Which fastpath did you find this useful for?

The one you added for populating advance widths: https://github.com/harfbuzz/harfrust/blame/d9ca71377d0ab07e805dcd724412d79575f4269e/harfrust/src/hb/glyph_metrics.rs#L114

We can't really fall back to this without checking because a default FontFuncs::populate_advance_widths _should_ call FontFuncs::advance_width in case the user provided that one...

harfrust/src/hb/font_funcs.rs
dfrg: The `builtin` object exists precisely so that a user can choose which functions to override and leave the others to default implementations. Since traits in rust can't carry state, we have to pass it as a parameter and make calls into that in the provided methods.

re parenting: with the way this works, the user would have to pass in the entire delegation chain along with their impl of `FontFuncs`. Seems simpler to just allow them to handle inheritance/delegation _within_ their own `FontFuncs` impl and then we don't have to worry about lifetimes or resource management.

harfrust/src/hb/font_funcs.rs
dfrg: Because as you pointed out, it's simply incorrect :) Removed this and changed `vertical_origin` to return both coordinates, with the first defaulting to `x_advance / 2`.

Chrome doesn't override `glyph_h_origin`, so I'll be leaving it out for now.

behdad: > Chrome doesn't override `glyph_h_origin`, so I'll be leaving it out for now.

Yeah non-0 h-origin is not used by any client. It was a speculative feature in HB.

PR #368, opened 2026-05-26
0.8.0

drott:

- Refactor ot shape context functions into methods.
- Added `ShapeOptions` type as argument to shape calls, preparation for being able to pass font functions. `ShapeOptions` accepts shape plan, point size, features. `shape_with_plan()` function removed in favor of new builder-style API.
- Document release process to avoid missing LICENSE files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.8.0] - 2026-05-26
+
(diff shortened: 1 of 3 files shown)


behdad:
> * Added ShapeOptions type as argument to shape calls, preparation for being able to pass font functions. ShapeOptions accepts shape plan, point size, features. shape_with_plan() function removed in favor of new builder-style API.

For the future, you should feel free to use back-tick quoting the symbol names.

drott:
> For the future, you should feel free to use back-tick quoting the symbol names.

Fixed and force-pushed.

PR #371, opened 2026-05-29
Add codepoint-based context setters to UnicodeBuffer

drott:
Allow setting pre- and post-context from raw u32 codepoints directly, avoiding the need to round-trip through UTF-8 strings for users who already have codepoint representation (e.g. integrations with C++ HarfBuzz).

Assisted-by: Google Antigravity

### DIFF
--- a/harfrust/src/hb/buffer.rs
+++ b/harfrust/src/hb/buffer.rs
@@ -1601,10 +1601,26 @@ impl hb_buffer_t {
         }
     }
 
+    fn set_pre_context_codepoints(&mut self, codepoints: &[u32]) {
+        self.clear_context(0);
+        for (i, &c) in codepoints.iter().take(CONTEXT_LENGTH).enumerate() {
+            self.context[0][i] = c;
+            self.context_len[0] += 1;
+        }
+    }
+
     fn set_post_context(&mut self, text: &str) {
         self.clear_context(1);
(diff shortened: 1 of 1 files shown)


harfrust/src/hb/buffer.rs
```diff
@@ -1773,12 +1789,24 @@ impl UnicodeBuffer {
         self.0.set_pre_context(str);
     }
 
+    /// Sets the pre-context for this buffer from codepoints.
+    #[inline]
+    pub fn set_pre_context_codepoints(&mut self, codepoints: &[u32]) {
```
dfrg, line 1798: It looks like this expects HB’s internal representation of pre-context. That is, the sequence of code points prior to the shaping text in reverse order. I’m not opposed to this API but we should at least document the expected ordering.

dfrg (approved):
One nit about adding some clarity to docs but otherwise LGTM. Thanks!

behdad:
LGTM with Chad's comment.

harfrust/src/hb/buffer.rs
drott: Ah yes, I missed the `.rev()` in the `str` based pre context function. Will document.

drott: Done.

PR #373, opened 2026-05-30
Add Language::new that takes byteables

behdad:
We should probably change the uses of `Language::from_str()` in `tag.rs` to use this as well.

### DIFF
--- a/harfrust/src/hb/common.rs
+++ b/harfrust/src/hb/common.rs
@@ -194,6 +194,12 @@ type SmallVecLanguage = SmallVec<[u8; 8]>;
 pub struct Language(SmallVecLanguage);
 
 impl Language {
+    /// Creates a new language from the given bytes.
+    #[inline]
+    pub fn new(bytes: impl AsRef<[u8]>) -> Self {
+        Language::from_bytes(bytes.as_ref())
+    }
+
     /// Returns the language as a string.
     #[inline]
     pub fn as_str(&self) -> &str {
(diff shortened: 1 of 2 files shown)


harfrust/src/hb/common.rs
```diff
@@ -194,6 +194,12 @@ type SmallVecLanguage = SmallVec<[u8; 8]>;
 pub struct Language(SmallVecLanguage);
 
 impl Language {
+    /// Creates a new language from the given bytes.
+    #[inline]
+    pub fn new(bytes: impl AsRef<[u8]>) -> Self {
```
dfrg: The current `from_str` conversion fails if the string is empty. Should we match that behavior and make this return `Option::None`? What does HB do?

One consideration is that we deal with `Option<Language>` in a lot of places and it would be unfortunate if `Some(Language::new(“”))` and `None` were treated differently.

harfrust/src/hb/common.rs
```diff
@@ -194,6 +194,12 @@ type SmallVecLanguage = SmallVec<[u8; 8]>;
...
+        Language::from_bytes(bytes.as_ref())
+    }
+
     /// Returns the language as a string.
     #[inline]
     pub fn as_str(&self) -> &str {
```
dfrg, line 212: It would be useful to have a public `as_bytes(&self) -> &[u8]` as well.

dfrg (approved):
Thanks for adding this. Generally LGTM with a question about handling empty languages.

harfrust/src/hb/common.rs
behdad: `Option` sounds right. HB returns `HB_LANGUAGE_INVALID` on empty string.

PR #374, opened 2026-05-30
Re-export GlyphId

dfrg:
This is now used in `FontFuncs` and currently requires the user to add an explicit `read-fonts` or `font-types` dependency to access it.


nicoburns (approved):
+1. I ran into this when upgrading Servo.

PR #378, opened 2026-06-01
Add bulk UnicodeBuffer glyph-info append

behdad:
Expose a UnicodeBuffer method that appends a slice of GlyphInfo values after reserving once. This lets embedding bridges populate buffers from layout-compatible glyph-info arrays without per-codepoint API overhead.

Testing: cargo check --manifest-path harfrust/Cargo.toml

Assisted-by: OpenAI Codex

### DIFF
--- a/harfrust/src/hb/buffer.rs
+++ b/harfrust/src/hb/buffer.rs
@@ -1894,6 +1894,19 @@ impl UnicodeBuffer {
         self.0.set_post_context_codepoints(codepoints);
     }
 
+    /// Appends glyph infos to a buffer.
+    #[inline]
+    pub fn push_glyph_infos(&mut self, infos: &[GlyphInfo]) -> bool {
+        let len = infos.len();
+        if !self.0.ensure(self.0.len + len) {
+            return false;
+        }
+
+        self.0.info[self.0.len..self.0.len + len].copy_from_slice(infos);
+        self.0.len += len;
+        true
+    }
+
     /// Appends a character to a buffer with the given cluster value.
     #[inline]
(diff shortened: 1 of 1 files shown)


harfrust/src/hb/buffer.rs
```diff
@@ -1894,6 +1894,20 @@ impl UnicodeBuffer {
         self.0.set_post_context_codepoints(codepoints);
     }
 
+    /// Appends glyph infos to a buffer.
+    #[inline]
+    pub fn add_from_glyph_infos(&mut self, infos: &[GlyphInfo]) -> bool {
```
dfrg: Let’s call this `push_glyph_infos` just to align better with `push_str` and more idiomatic rust naming in general.

harfrust/src/hb/buffer.rs
```diff
@@ -1894,6 +1894,20 @@ impl UnicodeBuffer {
...
+        let len = infos.len();
+        if !self.0.ensure(self.0.len + len) {
+            return false;
+        }
+
+        self.0.info[self.0.len..self.0.len + len].copy_from_slice(infos);
+        self.0.len += len;
+        self.0.context_len[1] = 0;
```
dfrg: Do we want to reset the post context here? The `push_str` method doesn’t and we probably want to remain consistent.

dfrg (approved):
Generally LGTM with a couple nits.

harfrust/src/hb/buffer.rs
behdad: Agreed. In HB, pre/post-context are automatic. In HR, they aren't.

PR #380, opened 2026-06-02
Use language bytes in tag parsing

behdad:


harfrust/src/hb/tag_table.rs
```diff
@@ -3,1675 +3,1675 @@
...
-    LangTag { language: "zu", 	tag: Tag::new(b"ZUL ") }, // Zulu
-    LangTag { language: "zum", 	tag: Tag::new(b"LRC ") }, // Kumzari -> Luri
-    LangTag { language: "zyb", 	tag: Tag::new(b"ZHA ") }, // Yongbei Zhuang -> Zhuang
-    LangTag { language: "zyg", 	tag: Tag::new(b"ZHA ") }, // Yang Zhuang -> Zhuang
-    LangTag { language: "zyj", 	tag: Tag::new(b"ZHA ") }, // Youjiang Zhuang -> Zhuang
-    LangTag { language: "zyn", 	tag: Tag::new(b"ZHA ") }, // Yongnan Zhuang -> Zhuang
-    LangTag { language: "zyp", 	tag: Tag::new(b"QIN ") }, // Zyphe Chin -> Chin
-//  LangTag { language: "zza", 	tag: Tag::new(b"ZZA ") }, // Zazaki [macrolanguage]
-    LangTag { language: "zzj", 	tag: Tag::new(b"ZHA ") }, // Zuojiang Zhuang -> Zhuang
+    LangTag { language: *b"aa\x00\x00", 	tag: Tag::new(b"AFR ") }, // Afar
```
drott: Having the explicit trailing null bytes here looks a bit noisy in the code. Can we fix that with a macro, or something like a compile time constructor for language that does the padding?

harfrust/src/hb/tag_table.rs
```diff
@@ -3,1675 +3,1675 @@
 use read_fonts::types::Tag;
 
 pub(crate) struct LangTag {
-    pub language: &'static str,
+    pub language: [u8; 4],
```
drott, line 6: Yes, good to see this bit.

drott (approved):
LGTM, but would have a slight preference for avoiding the visually noisy `\x00` padding in `tag_table.rs`.

harfrust/src/hb/tag_table.rs
behdad: Agreed. Lemme see.

drott (approved):
LGTM, thanks.

PR #381, opened 2026-06-02
0.8.4

behdad:

- Use bytes instead of str storage for language tag table.


drott:
> * Use bytes instead of str storage for language tag table.

LGTM under the assumption this is not affecting public API, which I don't think it is.

drott (approved):

PR #383, opened 2026-06-11
POC: parse context rules in a single pass in the apply path

cmyr:
In the course of investigating some other ideas claude suggested an optimization unrelated to fontations, and I had it draft a POC and do some simple benchmarking. It looks to offer a meaningful win for nastaliq, so I am sharing this for discussion.

Handing it off to claude:

---------

In the contextual lookup apply path, rule fields (backtrack / input / lookahead / lookup records) were fetched through the generated lazy getters. Each getter re-derives the byte ranges of all preceding fields, re-reading the leading counts every time, so examining and then applying a candidate chain rule touched the leading counts ~15 times where four reads suffice.

Two changes, which only pay off together:

1. ParsedRule: read each examined rule's counts once, in order, and store the four fields as slices. This removes the per-getter re-derivation for rules that reach the match/apply path. Glyph and class sequences are both read as raw u16s, which lets one type serve all four rule flavors (replacing the ContextRule/ChainContextRule getter traits and ToU16).

2. ContextRule::first_input: a two-scalar probe used by the skip-ahead loops. Skip-ahead visits (and discards) most rules in a large rule set and only needs the first input glyph/class; parsing the whole rule there (change 1 alone) costs more than the old lazy getters, which gave first-input access in two reads.

No unsafe; no intended behavior change (all shaping tests pass).

Interleaved A/B benchmarks vs. main (3 rounds, means):

  NotoNastaliqUrdu/fa-thelittleprince  55.64ms -> 51.22ms  (-7.9%)
  NotoNastaliqUrdu/fa-words            59.63ms -> 54.75ms  (-8.2%)
  Amiri, Roboto, Devanagari, SourceSerif: within noise (<= 1%)

The same patch on the sanitize experiment branch (bench-sanitize-v3, where offset resolution is already cheaper) measured -5.1% / -6.3% on Nastaliq, so the two optimizations compose rather than overlap.

The same idea could instead live in read-fonts: codegen could emit eager read impls for the var-size rule tables. Doing it in harfrust keeps the experiment local.

### DIFF
--- a/harfrust/src/hb/ot/contextual.rs
+++ b/harfrust/src/hb/ot/contextual.rs
@@ -14,8 +14,8 @@ use read_fonts::tables::layout::{
     ChainedSequenceContextFormat3, ChainedSequenceRule, ClassSequenceRule, SequenceContextFormat1,
     SequenceContextFormat2, SequenceContextFormat3, SequenceLookupRecord, SequenceRule,
 };
-use read_fonts::types::{BigEndian, GlyphId, GlyphId16, Offset16};
-use read_fonts::{ArrayOfOffsets, FontRead};
+use read_fonts::types::{BigEndian, FixedSize, GlyphId, Offset16};
+use read_fonts::{ArrayOfOffsets, FontData, FontRead};
 
 impl WouldApply for SequenceContextFormat1<'_> {
     fn would_apply(&self, ctx: &WouldApplyContext) -> bool {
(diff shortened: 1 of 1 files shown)


behdad:
This is legit indeed. I defer to Chad for technicalities.

dfrg:
This is simpler and faster with the caveat that we do a little bit of manual parsing which certainly has precedent here. I’m in favor of landing this.

behdad:
This speeds up duployan benchmark by 20%. I'm merging.

Issue #384, opened 2026-06-12
Add overflow panic with Blink's `HarfBuzzShaperTest.OverlyLongGraphemeCluster`


drott:
Shaping an "overly long" grapheme cluster 

```
 // Letter 'e' with 35000 diacritics, followed by letter 'X'
  StringBuilder builder;
  builder.Append('e');
  for (unsigned i = 0; i < 35000; ++i)
    builder.Append(uchar::kCombiningAcuteAccent);
  builder.Append('X');
  String string = builder.ToString();
```

leads to a `panic_const_add_overflow`.

```
  0000000008f61245  core::panicking::panic_const::panic_const_add_overflow                            core.a3ab07c529f88899-cgu.08:0:0
  000000000914b08b  harfrust::hb::ot_layout_gpos_table::propagate_attachment_offsets                  harfrust.3f603180a8391e7a-cgu.13:0:0
  000000000914ce49  harfrust::hb::ot_layout_gpos_table::GPOS::position_finish_offsets                 harfrust.3f603180a8391e7a-cgu.13:0:0
  000000000911cd35  <harfrust::hb::ot_shape::OtShapeContext>::shape_internal                          harfrust.3f603180a8391e7a-cgu.06:0:0
  0000000009113926  <harfrust::hb::face::hb_font_t>::shape_with_plan                                  harfrust.3f603180a8391e7a-cgu.05:0:0
  0000000009113a4f  <harfrust::hb::face::hb_font_t>::shape                                            harfrust.3f603180a8391e7a-cgu.05:0:0
  00000000090e14a7  _hb_harfrust_shape_rs                                                             harfbuzz_rust.48d6ac38f09ebe87-cgu.0:0:0
  0000000007a9e297  _hb_harfrust_shape                                                                hb-harfrust.cc:0:0
  0000000007b058bf  hb_shape_plan_execute                                                             hb-shape-plan.cc:0:0
  0000000007b05daa  hb_shape_full                                                                     hb-shape.cc:0:0
```

behdad:
That's probably the spatial offsets overflowing. We can use saturating math.

drott:
This is unfortunately not fixed in 0.10.0.

behdad:
Do you have a new stacktrace?

drott:
Yes, I was just looking at that, I think we have to move it to HarfBuzz - 

```
[...]

PR #387, opened 2026-06-15
Release 0.8.5

drott:

- Fix GPOS attachment offset overflow #384
- Require read-fonts 0.40.


dfrg:
Updating read-fonts to a new minor version is a breaking change here so this needs to be 0.9.0. Otherwise LGTM.

dfrg:
> Updating read-fonts to a new minor version is a breaking change here so this needs to be 0.9.0. Otherwise LGTM.

This should probably be noted in the release instructions.

PR #389, opened 2026-06-16
Adopt new read-fonts API

dfrg:
Add support for the new read-fonts API behind the `experimental_font_api` feature. Requires googlefonts/fontations#1852 to land and be released first.

Unfortunately a lot of churn and some mess here in order to maintain the current API while we wait for the read-fonts code to stabilize.


harfrust/src/hb/face.rs
```diff
@@ -387,19 +412,99 @@ impl Scale {
...
+) -> GlyphBuffer {
+    let Some(hb_font) = hb_font_t::from_font(font) else {
+        buffer.clear();
+        return GlyphBuffer(buffer.0);
+    };
+    // If the user didn't request an explicit scale but the font instance
+    // has a size, set the scale to that size with 16 fractional bits.
+    if options.scale.is_none() {
+        if let Some(ppem) = font.size() {
+            options = options.scale(Some((ppem * 65536.0) as i32));
```
behdad, line 440: This is concerning to me. In HB we default to upem for scale.

dfrg: Yeah, I meant to call this one out to you. I feel like we should do something to make pure Rust use “just work” with the configured instance from fontations. I’m just not sure what.

behdad: Ultimately up to you.

16bit fractional may also be too much, leaving little to the integral part, in case someone decides to accumulate glyph positions in the same space, and for a large enough font size.

Does skrifa expose any 26.6 at all? I feel like that's a better default if you really want to pick one.

PR #397, opened 2026-06-24
Fix overflow panic in skipping_iterator_t with explicit saturating_add

drott:
When shaping extremely long lines with fonts that have classic `kern` or
AAT `kerx` tables, the shaper instantiates a single `skipping_iterator_t`
once outside the loop and reuses it across the entire buffer. Because
`self.glyph_data` accumulates matching sequence indices without being reset,
it can eventually exceed `u16::MAX` (65,535) and cause an addition
overflow panic.

This fix upgrades `self.glyph_data` to a `u32`, and propagates the type
upgrade to match_func.

Assisted-by: Gemini

### DIFF
--- a/harfrust/src/hb/ot/contextual.rs
+++ b/harfrust/src/hb/ot/contextual.rs
@@ -37,7 +37,7 @@ impl WouldApply for SequenceContextFormat1<'_> {
                                     glyph_id: ctx.glyphs[i + 1].into(),
                                     ..GlyphInfo::default()
                                 };
-                                match_glyph(&mut info, value.get().to_u16())
+                                match_glyph(&mut info, value.get().to_u32())
                             })
                     })
                 })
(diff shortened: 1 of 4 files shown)


drott:
Not 100% sure if this is the right approach,  but does help against issues with encountering overflows with AAT font shaping.

behdad:
If I understand the code correctly, this should simply be a uint and no saturation needed. In fact, the fact that it's 16bit is one of those beefs I have with this codebase baking a lot of gid16 assumptions around.

drott:
I'd prefer to avoid machine-sized types like `std::os::raw::c_uint` and instead have a specific size, but do you mean `u64` then or `u32` as I did in this change?

behdad:
`u32` is fine, but I prefer to change the`may_match` and other signatures to pass down u32 as well, all the way. Should need no saturation.

dfrg:
> `u32` is fine, but I prefer to change the`may_match` and other signatures to pass down u32 as well, all the way. Should need no saturation.

+1. Use of unnecessarily narrow types is endemic in this codebase and choosing promotion rather than truncation has been my strategy when I encounter these.

drott:
Upgraded to u32 and propagating this to `match_func` PTAL.

dfrg (approved):
LGTM, thanks!

PR #399, opened 2026-06-29
Fix custom nominal glyph cmap caching

### LINKED ISSUE #396: Suspected shape plan caching issue leading to missed call to nominal_glyph hb_font function
Testing HB-HR and HarfRust 0.10.0 in Chrome we get a failure in unicode-range tests:

Failures: https://ci.chromium.org/ui/p/chromium/builders/try/linux-rel/b8678201834162803169/test-results?q=ExactID%3A%3A%2F%2F%5C%3Ablink_web_tests%21webtest%3A%3Ainspector-protocol%2Flayout-fonts%23unicode-range-order.js+VHash%3A32f081785efe9631&clean= 

Test source is https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/inspector-protocol/layout-fonts/unicode-range-order.js;l=97?q=unicode-range-order.js - test3 behaves like test2 even though test3 has a different unicode-range restriction.

Observation: test3 behaves just like test2 even though it should behave differently, as on test3 fonts the unicode-range setting is different. This works with HarfBuzz built-in OT shaper but does not work with HarfRust. 

When a nominal glyph function is set, glyph coverage cannot be assumed to stay identical between calls.

behdad:
Fixes https://github.com/harfbuzz/harfrust/issues/396

### DIFF
--- a/harfrust/src/hb/font_funcs.rs
+++ b/harfrust/src/hb/font_funcs.rs
@@ -312,17 +312,12 @@ impl<'a, 'u> FontFuncsDispatch<'a, 'u> {
 
     #[inline(always)]
     pub(crate) fn nominal_glyph(&mut self, c: u32) -> Option<GlyphId> {
-        let cache = self.builtin.face.cmap_cache;
-        if let Some(gid) = cache.get(c) {
+        if let Some(funcs) = &mut self.funcs {
+            funcs.nominal_glyph(&self.builtin, c)
+        } else if let Some(gid) = self.builtin.face.cmap_cache.get(c) {
             Some(gid.into())
-        } else if let Some(funcs) = &mut self.funcs {
-            if let Some(gid) = funcs.nominal_glyph(&self.builtin, c) {
-                cache.set(c, gid.to_u32());
-                Some(gid)
-            } else {
-                None
-            }
         } else if let Some(gid) = self.builtin.nominal_glyph(c) {
+            let cache = self.builtin.face.cmap_cache;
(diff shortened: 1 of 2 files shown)


drott:
LGTM, thanks, then we can still add that to the release.

drott:
🤔 The Blink issue still does not seem to be fixed and still picks up info from the previous font object:

https://results.usercontent.luci.app/invocations/task-chromium-swarm.appspot.com-790d5f69400e2f11/tests/:%2F%2F%5C:blink_web_tests%21webtest::inspector-protocol%2Flayout-fonts%23unicode-range-order.js/results/00ef55d3-07160/artifacts/pretty_text_diff?token=AXsiX2kiOiIxNzgyODk0ODI1MTM0IiwiX3giOiIzNjAwMDAwIn0gB8vlxZHH02fcXNv6MYizq-8bGtQR3uQnu3zUpRQU7w

image

behdad:
> 🤔 The Blink issue still does not seem to be fixed and still picks up info from the previous font object:

I'm thinking something else is going on here, not HR-side caching. Let me assess.

drott:
Scratch my previous comment, they work on Mac, but they still fail on Linux. 

https://crrev.com/c/7912622/7?checksRunsSelected=linux-rel&tab=checks

drott:
On the latest run and locally tested, this is no longer reproducible on Linux either. I am not sure what happened on the bot there. 
https://crrev.com/c/7912622?checksPatchset=8&tab=checks

PR #407, opened 2026-07-08
Add optional icu4x unicode data

dfrg:
Adds a new `icu` feature that pulls in the `icu_properties` and `icu_normalizer` crates to provide Unicode support for general category, script, canonical combining class, mirroring, compose and decompose (the HB unicode funcs set).

When the feature is enabled, we use the ICU functions and our built-in UCD is excluded from the build.

Also includes a little bit of reorganization and cleanup of the unicode types.


harfrust/src/hb/algs.rs
```diff
@@ -14,12 +14,6 @@ pub fn rb_flag_unsafe(x: u32) -> u32 {
     }
 }
 
-// FLAG_RANGE macro in harfbuzz.
-#[inline]
-pub fn rb_flag_range(x: u32, y: u32) -> u32 {
-    (x < y) as u32 + rb_flag(y + 1) - rb_flag(x)
-}
```
dfrg, line 21: This port from HB of `static_assert_expr((x) < (y)) + FLAG((y) + 1) - FLAG(x)` wasn't quite right :)

behdad (approved):

PR #421, opened 2026-08-20
Cache GDEF mark filtering sets in bitmaps

behdad:
## Summary

- port HarfBuzz PR #6188 to the Rust GDEF cache
- flatten narrow mark-filtering sets into exact biased 512-glyph bitmaps
- retain the digest and out-of-line table fallback for wider sets

Upstream: https://github.com/harfbuzz/harfbuzz/pull/6188

## Performance

Noto Sans Duployan with the Duployan benchmark corpus improves by a confirmed 10%.

In the local 10-repetition run, median CPU time decreased from 2.137 s to 1.893 s and median wall time decreased from 2.146 s to 1.901 s (11.4%).

## Validation

- cargo check --workspace --all-features
- cargo test -p harfrust --lib (210 passed)
- cargo clippy -p harfrust --all-features -- -D warnings
- HarfBuzz test-shape-harfrust (4 subtests passed)
- git diff --check

### DIFF
--- a/harfrust/src/hb/ot/mod.rs
+++ b/harfrust/src/hb/ot/mod.rs
@@ -42,6 +42,32 @@ pub enum MarkGlyphSetBitmap {
 }
 
 impl MarkGlyphSetBitmap {
+    fn page_from_coverage(coverage: &CoverageTable, min_gid: u32) -> Option<[u64; 8]> {
+        let mut bits = [0; 8];
+        let mut add_glyph = |glyph: u32| {
+            let biased = glyph.wrapping_sub(min_gid);
+            *bits.get_mut((biased / 64) as usize)? |= 1 << (biased & 63);
+            Some(())
+        };
+
+        match coverage {
+            CoverageTable::Format1(table) => {
+                for glyph in table.glyph_array() {
+                    add_glyph(u32::from(glyph.get()))?;
+                }
+            }
+            CoverageTable::Format2(table) => {
+                for range in table.range_records() {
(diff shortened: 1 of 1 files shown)


behdad:
@dfrg Looks like bots are failing for icu4x related reasons.

harfrust/src/hb/ot/mod.rs
```diff
@@ -30,7 +30,81 @@ pub struct OtCache {
...
+                }),
+        };
+
+        if let Some((min_gid, max_gid)) = bounds {
+            if max_gid.wrapping_sub(min_gid) < MARK_GLYPH_SET_PAGE_BITS {
+                let mut bits = [0; 8];
+                match coverage {
+                    CoverageTable::Format1(table) => {
+                        for glyph in table.glyph_array() {
+                            let biased = u32::from(glyph.get()) - min_gid;
```
dfrg: Since we only use the first and last values from the coverage table, we can't guarantee that `glyph >= min_gid` here and this could panic with overflow. We also don't check the upper bound so we might panic on the index. The same applies for format 2 below.

I'd recommend pulling this logic into a separate function (or an inlined closure) that returns `Option<[u64; 8]>`. Then use wrapping sub to compute the biased index and `*bits.get_mut((biased / 64) as usize)? |= ...` which should catch both boundary conditions?

dfrg (approved):
LGTM once we address the potential panics for unsorted coverage tables.

I'll send another PR that silences the ICU warnings and we can follow up with a comment on the related issue.

behdad:
@dfrg Thanks. PTAL.

dfrg:
Looks good, thanks!

PR #423, opened 2026-08-20
Hot-loop discipline: block copies, slice scans, clone-free recursion

behdad:
Three stacked, independently-tested commits that tighten the hottest loops. Each passes the full test suite (6,267 tests) on its own.

## Changes

1. **[buffer] Replace per-element copy loops with block copies** — `move_to` (both branches), `shift_forward`, `make_room_for`, `next_glyphs`, and `replace_glyphs` copied glyph infos one element at a time, re-testing `have_separate_output`, redoing the bytemuck cast, and paying bounds checks per element. Now they use `copy_within`/`copy_from_slice`, matching the `memmove`s HarfBuzz uses.

   **This is also a semantic fix**: hb uses `memmove` in `move_to`'s rewind branch (hb-buffer.cc). When `info`/`out_info` share the array and a deletion gap is smaller than the rewind distance, the old ascending element copy re-read slots it had already overwritten, smearing glyphs. `copy_within` restores exact hb semantics.

2. **[layout] Scan for lookup candidates over slices in apply loops** — `apply_forward`/`apply_backward` now hoist `face`/`lookup_mask`/`lookup_props` into locals (sound: `recurse` saves and restores them) and scan with `position`/`rposition` over a slice, removing the per-glyph bounds check from the digest/mask/props test.

3. **[gsubgpos] Move match_positions on recursion instead of cloning** — `recurse()` cloned the `match_positions` SmallVec on every nested-lookup application; the child never reads the parent's positions, so it now gets a fresh inline SmallVec (`from_elem(0, 1)`, preserving the len ≥ 1 invariant that `match_input`'s single-glyph fast path relies on) and the parent's is restored by move. No allocation, no copy.

## Benchmarks

Instruction counts (perf stat on a corpus driver mirroring benches/shaping.rs; per-pass counts via a 10-rep/20-rep subtraction so startup and one-time cache building cancel; ~0.01% measurement variance):

| Target | Δ instructions/pass |
|---|---|
| Amiri / fa-thelittleprince | −4.5% |
| NotoNastaliqUrdu / fa-thelittleprince | −2.4% |
| Roboto / en-thelittleprince | −2.4% |
| SourceSerifVariable / react-dom | −1.8% |
| NotoNastaliqUrdu / fa-words | −0.7% |
| NotoSansDevanagari / hi-words | −0.3% |
| Roboto / en-words | +0.0% |

Wall-clock criterion A/B showed larger per-target numbers (−2% to −6%) but those fold code-layout noise into the delta; the instruction counts above are the honest measurement. Instruction-count bisection also caught (and led to fixing) an inlining regression in an earlier formulation of the layout commit: an Iterator::position-based scan grew apply_forward enough to shift LLVM's inlining in the apply path, costing +0.8% on Roboto/en-words; the manual-index-loop form in the final commit keeps the bounds-check elision without perturbing inlining.

### DIFF
--- a/harfrust/src/hb/buffer.rs
+++ b/harfrust/src/hb/buffer.rs
@@ -627,6 +627,29 @@ impl hb_buffer_t {
         self.out_info_mut()[i] = info;
     }
 
+    /// Block-copies `info[src..src + count]` to `out_info[dst..dst + count]`.
+    /// The caller must have made room in the out buffer already.
+    #[inline]
+    fn copy_infos_to_out(&mut self, src: usize, dst: usize, count: usize) {
+        if self.have_separate_output {
+            let out: &mut [GlyphInfo] = bytemuck::cast_slice_mut(self.pos.as_mut_slice());
+            out[dst..dst + count].copy_from_slice(&self.info[src..src + count]);
+        } else {
+            self.info.copy_within(src..src + count, dst);
+        }
+    }
+
+    /// Block-copies `out_info[src..src + count]` to `info[dst..dst + count]`.
+    #[inline]
+    fn copy_out_to_infos(&mut self, src: usize, dst: usize, count: usize) {
+        if self.have_separate_output {
+            let out: &[GlyphInfo] = bytemuck::cast_slice(self.pos.as_slice());
+            self.info[dst..dst + count].copy_from_slice(&out[src..src + count]);
+        } else {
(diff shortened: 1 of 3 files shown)


dfrg (approved):
Love this.

I did try some of these independently (things like copy_from_slice and copy_within) without seeing gains, but I guess they work better in combination with other changes. The loop hoisting is particularly interesting.

Ship it!

PR #424, opened 2026-08-20
Defer context rule parsing behind input probes

behdad:
Stacked on #423. Two commits, each independently measured with instruction counts (perf stat on a corpus driver, 10/20-rep subtraction, ~0.01% variance) and independently green on the full test suite.

Class-based contextual lookups are where harfrust trails HarfBuzz the most: same-binary `hb-shape --shaper ot` vs `--shaper harfrust` measures NotoNastaliqUrdu/fa-words at roughly +36% instructions before this work. Both rule loops parsed every visited rule (`ParsedRule` re-reads five counts and validates four arrays) before rejecting most rules on the first one or two input values.

1. **Plain rule loop**: probe the first/second input values straight from rule data (`first_input`/new `second_input`) and parse only survivors. NotoNastaliqUrdu: **−7.4%** (fa-thelittleprince), **−6.6%** (fa-words); other targets unchanged.
2. **Chained rule loop**: same idea via a `ChainRuleProbe` that reads the backtrack/input counts once and serves input/lookahead values on demand (short-input rules pre-match on lookahead values, per the existing logic). On top of commit 1: **−1.5% / −1.8%** on NotoNastaliqUrdu; other targets unchanged.

Also tried and rejected by measurement, for the record: eager per-phase coverage resolution in the Format3 appliers (+10.7% on Amiri — those matchers are failure-dominated, so lazy per-position resolution is already right), and enabling the `rules.len() <= 4` fast-path TODO (+0.3% on NotoNastaliqUrdu — with commit 1 in place the probe loop beats the parse-everything path even for small rule sets, so the TODO comment's premise is now inverted).

### DIFF
--- a/harfrust/src/hb/ot/contextual.rs
+++ b/harfrust/src/hb/ot/contextual.rs
@@ -661,6 +661,75 @@ fn chain_rule_second_input(data: &FontData) -> Option<u16> {
     data.read_at(count_pos + 4).ok()
 }
 
+/// A chain rule with direct access to its raw data, for cheap positional
+/// probes without a full parse.
+trait ChainContextRule<'a>: ContextRule<'a> {
+    fn data(&self) -> FontData<'a>;
+}
+
+impl<'a> ChainContextRule<'a> for ChainedSequenceRule<'a> {
+    fn data(&self) -> FontData<'a> {
+        self.offset_data()
+    }
+}
+
+impl<'a> ChainContextRule<'a> for ChainedClassSequenceRule<'a> {
+    fn data(&self) -> FontData<'a> {
+        self.offset_data()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


behdad:
> 2\. **Chained rule loop**: same idea via a `ChainRuleProbe` that reads the backtrack/input counts once and serves input/lookahead values on demand (short-input rules pre-match on lookahead values, per the existing logic). On top of commit 1: **−1.5% / −1.8%** on NotoNastaliqUrdu; other targets unchanged.

I'm skeptical about this part, but part 1 sounds like a win.

dfrg (approved):
Code-wise, both commits LGTM.

The second one doesn't seem *un*reasonable so I'll defer to you whether you want to merge it.

PR #426, opened 2026-08-20
Copy small counts inline instead of calling memmove

behdad:
Fixes the AAT regression #423 introduced, visible in Behdad's daily tracking as Devanagari Sangam MN/hi-words dropping from ~20% slower than HarfBuzz to ~30% slower.

**Diagnosis.** The regression was invisible to instruction counts (bit-flat across #423) but obvious in cycles: **+8.8%** on Devanagari Sangam MN/hi-words with instructions unchanged. The block-copy conversion replaced inline element loops with `copy_within`/`copy_from_slice`, which lower to out-of-line `memmove` calls for runtime-variable counts. AAT morx state machines drive `move_to` with 1-3 glyph rewinds thousands of times, and at those sizes a memmove call costs roughly the same instructions as the old loop but far more cycles (call overhead plus size-dispatch branch misses). MORX-36 won 2× from the same conversion because its rewinds are ~359 glyphs — the opposite regime.

**Fix.** Counts below 16 copy inline; larger counts keep the block copy. The inline paths preserve the memmove overlap semantics #423 established: forward copies into out_info have `dst <= src` in the shared-array case (ascending copy is safe), while the rewind copy-back and `shift_forward` have `dst >= src` (descending copy).

**Validation** (perf stat, r=5):
- Devanagari Sangam MN/hi-words cycles: 3.46G → **3.19G**, matching the pre-#423 3.18G; instructions flat throughout.
- GeezaPro/fa-thelittleprince: measured unregressed locally in both instructions and cycles across #423 (its small move on the tracking chart looks like scatter).
- OT bench suite: instruction-identical to main on all seven targets.
- TestMORXThirtysix A×65536: 0.075s, unchanged — large moves stay on the block path.
- Full test suite green.

### DIFF
--- a/harfrust/src/hb/buffer.rs
+++ b/harfrust/src/hb/buffer.rs
@@ -627,26 +627,67 @@ impl hb_buffer_t {
         self.out_info_mut()[i] = info;
     }
 
+    /// Below this many elements, copy inline instead of calling memmove:
+    /// AAT state machines move a couple of glyphs at a time, and the call
+    /// plus memmove's size dispatch costs more cycles than the copy.
+    const SMALL_COPY: usize = 16;
+
     /// Block-copies `info[src..src + count]` to `out_info[dst..dst + count]`.
-    /// The caller must have made room in the out buffer already.
+    /// The caller must have made room in the out buffer already. In the
+    /// shared-array case the destination never starts past the source
+    /// (`dst <= src`), which makes the ascending inline copy overlap-safe.
     #[inline]
     fn copy_infos_to_out(&mut self, src: usize, dst: usize, count: usize) {
         if self.have_separate_output {
             let out: &mut [GlyphInfo] = bytemuck::cast_slice_mut(self.pos.as_mut_slice());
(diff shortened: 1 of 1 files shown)


behdad:
Threshold sweep (cycles, perf stat r=5), answering whether 16 is the right cutoff:

| SMALL_COPY | DevSangam/hi-words | Nastaliq/fa-prince | Roboto/en-prince |
|---|---|---|---|
| 4 | 3.178G | 3.078G | 598.9M |
| 8 | 3.182G | 3.040G | 602.0M |
| 16 | 3.195G | 3.078G | 604.2M |
| 32 | 3.198G | 3.084G | 601.7M |
| 64 | 3.175G | 3.076G | 601.0M |

Flat plateau: every value in 4-64 fully recovers the regression (main: 3.456G on DevSangam) and the residual spread is within code-layout noise. Keeping 16 as mid-plateau; the only hard constraints are staying above the 1-3 element AAT hot sizes and below large-rewind counts like MORX-36's ~359.

dfrg (approved):

PR #427, opened 2026-08-20
Update icu_properties usage

robertbastian:
Removes the use of deprecated `.to_icu4c_value()` APIs.

For `CanonicalCombiningClass`, the numerical value defined in Unicode is exposed in ICU4X 2.3, so this can just use that.

For `Script`, this removes the custom integer -> script mapping, and uses ICU4X's name API instead. Performance-wise, this is equivalent to the previous code: ICU4X stores names in a dense array over `[u8; 4]`, just like `SCRIPT_MAP` did. The conversion from ICU4X's script type through `[u8; 4]` to harfrust's script type is free.

### DIFF
--- a/harfrust/Cargo.toml
+++ b/harfrust/Cargo.toml
@@ -17,7 +17,7 @@ bitflags = "2.9"
 bytemuck = { version = "1.22", features = ["extern_crate_alloc"] }
 core_maths = { version = "0.1", optional = true }
 smallvec = "1.14"
-icu_properties = { version = "2.3.0", optional = true }
+icu_properties = { version = "=2.2.0", optional = true }
 icu_normalizer = { version = "2.2.0", optional = true }
 
 [features]
--- a/harfrust/Cargo.toml
+++ b/harfrust/Cargo.toml
@@ -17,7 +17,7 @@ bitflags = "2.9"
 bytemuck = { version = "1.22", features = ["extern_crate_alloc"] }
 core_maths = { version = "0.1", optional = true }
 smallvec = "1.14"
-icu_properties = { version = "=2.2.0", optional = true }
+icu_properties = { version = "2.3.0", optional = true }
 icu_normalizer = { version = "2.2.0", optional = true }
 
 [features]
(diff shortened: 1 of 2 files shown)


behdad:
Thank you. Can you address the CI rustfmt please?

dfrg:
The script changes LGTM. I noticed that chrome is still on 2.2.0 though so we might want to hold off on the version update and ccc change until chrome rolls icu_properties.

Thanks Robert.

robertbastian:
I've reverted CCC to use `to_icu4c_value` and the explicit 2.2 dependency.

robertbastian:
So there is a behaviour difference here: when ICU4X updates to a new Unicode version, it will return new scripts. The previous `SCRIPT_MAP` code would have returned `script::UNKNOWN` for these (until the map was updated), this code actually returns `Script`s with the new tags (e.g. `b"Seal"` for Unicode 18). It looks like this is fine, e.g. these will [fall through](https://github.com/harfbuzz/harfrust/blob/c400145b52f1e8b55053442add857f3d934ecd3a/harfrust/src/hb/ot_shaper.rs#L357) to the default shaper.

dfrg:
> So there is a behaviour difference here: when ICU4X updates to a new Unicode version, it will return new scripts. The previous `SCRIPT_MAP` code would have returned `script::UNKNOWN` for these (until the map was updated), this code actually returns `Script`s with the new tags (e.g. `b"Seal"` for Unicode 18). It looks like this is fine, e.g. these will [fall through](https://github.com/harfbuzz/harfrust/blob/c400145b52f1e8b55053442add857f3d934ecd3a/harfrust/src/hb/ot_shaper.rs#L357) to the default shaper.

This sounds like reduced maintenance burden to me :) Thanks again for helping to resolve this.

robertbastian:
@behdad should I try doing this in Harfbuzz as well?

behdad:
> @behdad should I try doing this in Harfbuzz as well?

I already pushed the same to HarfBuzz. Thanks.

PR #428, opened 2026-08-20
Walk morx subtables through packed descriptors with a digest gate

behdad:
Restructures the per-buffer morx walk:

- A flat, packed array of per-subtable descriptors (chain index, coverage, feature flags, data range, glyph-set digest) built once per face replaces re-walking the chain/subtable headers on every buffer. The whole filter walk for a 39-subtable font (Devanagari Sangam MN) touches ~2KB of contiguous, L1-resident memory.
- An O(1) `hb_set_digest_t` test gates each subtable before any `U32Set` work.
- The buffer's precise glyph set is built lazily, only once some subtable passes its digest.

Compiles and passes the full test suite against released read-fonts 0.43 (verified with the local path patch disabled), so this is mergeable now — but note the honest measurement: **standalone it is performance-neutral** in both the HarfBuzz parity harness and harfrust's own release profile. The walk savings are currently hidden under per-subtable accessor call overhead in read-fonts, and materialize with googlefonts/fontations#2061 (`#[inline]` on the AAT hot accessors): the combination measures **~15% faster** AAT shaping on Devanagari Sangam MN/hi-words (40.4ms → 34.4ms in HarfBuzz's `benchmark-shape`, taking harfrust from ~30% slower than HarfBuzz to ~5.5% on that benchmark, with byte-identical output).

A follow-up will switch per-application subtable kind resolution to the lifetime-free `SubtableParts` API proposed in googlefonts/fontations#2063 once it lands in a read-fonts release; googlefonts/fontations#2062 independently speeds up the legacy kern state machines on the same corpus.

### DIFF
--- a/harfrust/src/hb/aat/layout_morx_table.rs
+++ b/harfrust/src/hb/aat/layout_morx_table.rs
@@ -14,6 +14,7 @@ use read_fonts::tables::morx::{
     SubtableKind,
 };
 use read_fonts::types::{BigEndian, FixedSize, GlyphId16};
+use read_fonts::{FontData, FontRead};
 
 // Chain::compile_flags in harfbuzz
 pub fn compile_flags(face: &hb_font_t, builder: &AatMapBuilder, map: &mut AatMap) -> Option<()> {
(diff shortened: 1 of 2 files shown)


behdad:
Updated after probing Behdad's question about short-vs-long text: the digest machinery was indeed short-text-tuned. Instruction counts on Devanagari Sangam MN showed −3.4% for word-per-line shaping but +1.6% on synthetic 200-word lines. The branch now uses `hb_set_digest_t::full()` for buffers of ≥64 glyphs (a long run's digest saturates and rejects nothing anyway), which keeps the test branch-free — an explicit flag check costed ~3% by itself — and trims the long-line overhead to +1.4%, all of it distributed per-glyph digest maintenance. Typical text lines sit well under the threshold.

behdad:
Per Behdad's preference, the buffer/subtable digest gate is now dropped entirely — HarfBuzz's morx path has no digest either (the "set digest" comment in its chain accelerator refers to the zeroed glyph sets), so this restores strict design parity: eager buffer glyph set, bit-set intersection per subtable, and the packed descriptor walk as the only structural change. Numbers for the record: the digest carried about half of the short-text instruction win (−3.4% with, −1.6% without on Devanagari Sangam MN/hi-words) at the cost of a saturation threshold and per-mutation maintenance; it can be revisited with data if the combined stack ever justifies it.

dfrg (approved):
LGTM

PR #430, opened 2026-08-21
[aat] Add a fast-path to the state machine drivers

behdad:
Ports harfbuzz's AAT state-machine fast path (harfbuzz/harfbuzz#5667, Dec 2025 — landed after this driver was ported) to all three drive loops: morx `drive()`, kerx `apply_state_machine_kerning`, and the legacy kern machine.

When the machine transitions from start-of-text to start-of-text with no action while advancing — and the start state has no end-of-text action, precomputed as bit 0 of the existing `start_end_safe_to_break` mask — the driver consumes the whole run of same-class glyphs in a tight inner loop: per glyph it only calls `transition`, advances, and re-reads the (cached) class, skipping the state-entry lookup and the safe-to-break analysis (up to three entry lookups). Runs of glyphs the machine doesn't act on are the common case in real text.

Per-machine `max_ops` accounting is preserved exactly: morx charges only non-advancing transitions; kern/kerx charge every transition, fast path included.

Upstream this measured +12% on LucidaGrande and +6% on Menlo. Smoke check here: LucidaGrande/en-thelittleprince 5.97 ms → 5.49 ms (−8%) on the HarfBuzz parity harness, on top of the descriptor/parts stack. Outputs are byte-identical to the base branch on LucidaGrande, Menlo, GeezaPro, and Devanagari Sangam MN (full benchmark texts via hb-shape); full test suite passes.

Stacked on #428 and `perf/aat-subtable-parts` (which needs the now-merged googlefonts/fontations#2063 in a read-fonts release). Can be rebased to land separately if preferred — nothing here depends on the parts API.

### DIFF
--- a/harfrust/src/hb/aat/layout_kerx_table.rs
+++ b/harfrust/src/hb/aat/layout_kerx_table.rs
@@ -376,8 +376,11 @@ fn apply_state_machine_kerning<T, E, Driver: StateTableDriver<T, E>>(
     aat::StateEntry<E>: KerxStateEntryExt,
 {
     let mut state = START_OF_TEXT;
+    // Condition 3 below, precomputed for the start-of-text state: no
+    // end-of-text action can fire if we stop while in the start state.
+    let start_state_safe_to_break_eot = (c.start_end_safe_to_break & (1 << START_OF_TEXT)) != 0;
     c.buffer.idx = 0;
-    loop {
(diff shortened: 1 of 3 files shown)


behdad:
Depends on new read-fonts API, to be released tomorrow.

dfrg (approved):

PR #434, opened 2026-08-21
[font-funcs] Add a batched nominal-glyph callback

behdad:
Adds `FontFuncs::populate_nominal_glyphs`, mirroring the existing `populate_advance_widths` pattern: a `NominalGlyphBatch` view over a buffer run, with iterator access for safe implementations and a raw pointer/stride view (`RawNominalGlyphBatch`) for FFI bridges. The contract is deliberately identical to `hb_font_get_nominal_glyphs`: map consecutive codepoints until the first one the font has no glyph for, return the count.

The normalizer's short-circuit prefill — previously one `nominal_glyph` dispatch per character — now goes through the batch callback. The default trait implementation and the builtin path preserve the previous per-character behavior (including the cmap cache), so existing embedders see no change in behavior or performance.

The payoff is for embedders whose funcs cross an FFI boundary: HarfBuzz's harfrust shaper implements the callback with a single `hb_font_get_nominal_glyphs` call per run (hb branch `harfrust-batch-nominal`, PR to follow), replacing one round-trip per character. On the HarfBuzz parity benchmarks: **Menlo −6.2%, LucidaGrande −4.6%** (en-thelittleprince), GeezaPro −2%, Devanagari Sangam MN −1%. Outputs byte-identical on all four fonts; full test suite passes.

### DIFF
--- a/harfrust/src/hb/font_funcs.rs
+++ b/harfrust/src/hb/font_funcs.rs
@@ -106,6 +106,102 @@ impl<'a> IntoIterator for AdvanceWidthBatch<'a> {
     }
 }
 
+/// Raw C-style view over a batch of codepoints and output glyph ids.
+#[derive(Clone, Copy, Debug)]
+pub struct RawNominalGlyphBatch {
+    /// Number of batch entries.
+    pub len: usize,
+    /// Pointer to codepoints (read-only).
+    pub codepoints: *const u32,
+    /// Pointer to output glyph ids (writable).
+    pub glyphs: *mut u32,
+    /// Byte stride between successive codepoints.
+    pub codepoint_stride: isize,
+    /// Byte stride between successive glyphs.
+    pub glyph_stride: isize,
+}
+
+/// Safe batch view for codepoint to nominal-glyph mapping.
+///
+/// Glyph ids must be written for consecutive codepoints starting at the
+/// first entry; mapping stops at the first codepoint the font has no
+/// glyph for, and the number of glyphs written is returned from
+/// [FontFuncs::populate_nominal_glyphs].
+pub struct NominalGlyphBatch<'a> {
+    infos: &'a mut [GlyphInfo],
+}
+
+impl<'a> NominalGlyphBatch<'a> {
+    /// Byte offset of the output glyph id within a batch entry.
+    const GLYPH_OFFSET: usize = core::mem::offset_of!(GlyphInfo, vars)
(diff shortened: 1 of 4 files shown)


harfrust/src/hb/font_funcs.rs
```diff
@@ -106,6 +106,102 @@ impl<'a> IntoIterator for AdvanceWidthBatch<'a> {
...
+        }
+    }
+}
+
+pub struct NominalGlyphBatchIter<'a> {
+    infos: slice::IterMut<'a, GlyphInfo>,
+}
+
+impl<'a> Iterator for NominalGlyphBatchIter<'a> {
+    type Item = (u32, &'a mut u32);
```
dfrg: There’s a slight inconsistency here where the advance width batch iterator uses `GlyphId` (like all other `FontFuncs` functions) but this has `&mut u32`. I think bytemuck can handle the cast for us safely.

dfrg (approved):
LGTM with one nit.

harfrust/src/hb/font_funcs.rs
behdad: Good catch. A direct `&mut u32` → `&mut GlyphId` bytemuck cast doesn't compile against released font-types: `cast_mut` needs `NoUninit + AnyBitPattern` on both sides, and `GlyphId` only derives `AnyBitPattern` (only `Fixed`/`Point` got `NoUninit`). Rather than gate this PR on a font-types release, the iterator now yields a typed `NominalGlyphOut` slot with `set(GlyphId)`, so the public surface speaks `GlyphId` like the rest of `FontFuncs` and the raw view stays FFI-shaped. If font-types grows `NoUninit` on `GlyphId` later, this can become `&mut GlyphId` directly.

dfrg: ah, I missed that. PR for the fix in font-types: https://github.com/googlefonts/fontations/pull/2066

behdad: Now that main rolled to read-fonts 0.43.2 (font-types 0.12.4 with #2066's `NoUninit`), the branch is rebased and does exactly what you suggested: the iterator yields `(u32, &mut GlyphId)` via `bytemuck::cast_mut`, and the interim `NominalGlyphOut` slot is gone. Outputs re-verified byte-identical on the four AAT benchmark fonts.

dfrg: Thanks Claude Behdad!

PR #437, opened 2026-08-21
[ot] Cheapen the context rule walks

behdad:
Two measured reductions in the OT match machinery, aimed at the NotoNastaliqUrdu gap:

1. **Walk context rules on raw offsets, constructing rules only on probe pass.** The context and chain-context rule walks stepped an `ArrayOfOffsets` iterator, resolving each offset and constructing a rule table (with its read-time validation) for every visited rule — including the majority the first/second-input probes then discard. The walks now step the raw rule offset arrays and probe each rule's byte range directly; a rule is parsed only after its probes pass. The `ContextRule`/`ChainContextRule` traits, which existed solely to give the generic walk probe access, are deleted.

2. **Reuse the parsed base coverage in MarkBasePos apply** — the apply bound `base_coverage` at entry, then re-parsed it from scratch for the final base-index lookup (139K applies per pass on Nastaliq).

Measured (single-pass instruction counts; wall-clock improvement confirmed by Behdad):

| Font | instructions |
|---|---|
| NotoNastaliqUrdu/fa-thelittleprince | **−6.3%** (cycles −4.2%) |
| NotoSansDevanagari/hi-words | −0.2% |
| Amiri, Roboto | unchanged |

No new state anywhere — both changes only delete work. Outputs byte-identical to HarfBuzz on all four fonts; full test suite passes.

### DIFF
--- a/harfrust/src/hb/ot/gpos/mark.rs
+++ b/harfrust/src/hb/ot/gpos/mark.rs
@@ -148,7 +148,7 @@ impl Apply for MarkBasePosFormat1<'_> {
 
         // Checking that matched glyph is actually a base glyph by GDEF is too strong; disabled
         let base_glyph = info[idx as usize].as_glyph();
-        let Some(base_index) = self.base_coverage().ok()?.get(base_glyph) else {
+        let Some(base_index) = base_coverage.get(base_glyph) else {
             ctx.buffer
                 .unsafe_to_concat_from_outbuffer(Some(idx as usize), Some(ctx.buffer.idx + 1));
             return None;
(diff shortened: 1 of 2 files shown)


harfrust/src/hb/ot/gpos/mark.rs
```diff
@@ -148,7 +148,7 @@ impl Apply for MarkBasePosFormat1<'_> {
 
         // Checking that matched glyph is actually a base glyph by GDEF is too strong; disabled
         let base_glyph = info[idx as usize].as_glyph();
-        let Some(base_index) = self.base_coverage().ok()?.get(base_glyph) else {
+        let Some(base_index) = base_coverage.get(base_glyph) else {
```
dfrg, line 151: I think I've fixed this about five times but it never ended up in a branch worth submitting :)

dfrg (approved):
LGTM

harfrust/src/hb/ot/gpos/mark.rs
behdad: Yeah it wasn't worth anything but I told it to keep it.

PR #440, opened 2026-08-22
[kern] Decode Format1 state machines once per face

behdad:
Decode each legacy `kern` Format1 state machine into a flat transition array at subtable-cache build time: one packed `u32` per state-array byte, `[flags:16][safe:1][poison:1][new_state:14]`.

Everything the drive loop derived per transition — the `entry()` offset arithmetic, big-endian reads, new-state division and error plumbing, **plus the two extra `entry()` probes behind the safe-to-break computation** — depends only on `(state, class)`, so it is computed once per face and the hot loop is left with a single indexed load and bit tests. The decoded array is tiny (4 bytes per state-array byte) and strictly replaces reads of the equally-hot raw data, so this is not a footprint trade.

Semantics are preserved exactly, including error paths: the bounds-checked load reproduces `entry()`'s out-of-bounds → break behavior (unrepresentable new-state values clamp to a 0x3FFF sentinel row that always indexes past the array), and per-slot decode failures carry a poison bit. Machines with degenerate geometry (n_classes < 2, oversized state arrays) keep the raw `entry()` path.

Measurements (clang release rig, GeezaPro.ttc / fa-thelittleprince):
- instructions **168.5M → 160.5M (−4.8%)**, cycles **−5.5%**
- steady-state per-pass gap vs hb ot: **+16.2% → +10.6%** (154.5M vs 139.7M, `-n 10/20` subtraction)
- benchmark-shape wall: harfrust 10.3ms vs ot 9.20ms (**89.3%** of hb, from 85.5%)
- Devanagari Sangam MN: flat
- Outputs byte-identical to before on GeezaPro, Devanagari Sangam, Menlo, Lucida Grande; 6,267 tests pass; clippy `--all-features --all-targets -D warnings` clean

Also tried and rejected this session: bypassing the ClassCache for legacy kern in favor of the direct class-array read measured **+0.5% worse** (the 256-entry probe's branch-free hit path beats the array read's four data-dependent branches at kern's ~99% hit rate) — so the cache stays, on both engines.


behdad:
Second commit pushed: **[aat] Replace the buffer glyph set's U32Set with a flat bitmap**.

After the decoded machine landed the state machines at hb parity, the remaining GeezaPro excess was almost entirely the per-line buffer glyph set build: U32Set paid a page binary-search + checked indexing per insert (~95 instructions each). The new `GlyphSet` is a flat bitmap over the 16-bit glyph space — direct-indexed insert/contains, a deleted-glyph flag, and a touched-block mask so clears cost what the line touched. Subtable first/second sets use the same type, so `intersects` is a short word-wise AND. Bonus: the no_std `DigestU32Set` approximation is retired — no_std now gates exactly like std.

Cumulative measurements for the PR (clang release rig):

| Font | instructions vs main | gap vs hb ot (instr) | wall (benchmark-shape) |
|---|---|---|---|
| GeezaPro/fa | **−10.6%** | +16.2% → **+3.8%** | 9.49ms vs 9.03ms = **95.2%** (was 85.5%) |
| Devanagari Sangam/hi | **−6.4%** | **+12.6%** | 35.5ms vs 32.0ms = **90.1%** (was 86.3%) |
| Menlo/en | — | **+2.8%** | — |
| Lucida Grande/en | — | **−4.5% (harfrust ahead)** | — |

Outputs byte-identical on all four AAT fonts plus PingFang over a CJK sample; 6,267 tests; clippy clean; no_std+libm builds.

*Written by Claude (via Behdad's account).*

behdad:
Third commit pushed: **[morx] Decode extended state machines lazily, once per face** — the same treatment ported to the extended machines, with two twists learned the hard way:

1. **Lazy decode** (OnceLock, first post-gate application): Lucida Grande carries a 47K-cell contextual machine — eager decode charged every font instance ~16M instructions whether or not the text walks it. Lazy means unwalked machines cost zero and walked ones amortize.
2. **Formulation sensitivity**: several semantically-identical build implementations swung the release rig by up to ±19% through global LTO inlining shifts (one un-inlined `Lookup::value` across the whole morx path). The committed formulation is the one that measured best end-to-end; the commit message carries a warning for future refactors.

Cumulative PR standings (clang release rig, wall = benchmark-shape):

| Font | wall vs hb ot | day start |
|---|---|---|
| GeezaPro/fa | 9.44 vs 9.27ms = **98.2%** | 85.5% |
| Devanagari Sangam/hi | 33.5 vs 32.3ms = **96.4%** | 86.3% |
| Menlo/en | 3.60 vs 3.46ms = **96.1%** | 90% |
| Lucida Grande/en | 4.82 vs 5.02ms = **104% — ahead of harfbuzz** | 98.7% |

Instruction gaps vs hb: GeezaPro +2.5%, Sangam +4.3% steady-state; Lucida single-process runs carry the one-time decode (+12%) but win −10%/pass from the second shape on — the wall numbers above include that amortization.

Byte-identical outputs on all four AAT fonts + PingFang/CJK throughout; 6,267 tests; clippy clean; no_std+libm builds (no_std stays on the raw path).

*Written by Claude (via Behdad's account).*

behdad:
Fourth commit: **[kerx] Decode kerx state machines lazily, once per face** — same pattern as the morx commit, applied to the kerx Format1/Format4 drive loop (both share the u16 action-index payload, so one decoded type serves both).

Wall movement: **Menlo 96.1% → 98.3%**; Lucida Grande stays ahead of hb (4.74 vs 4.94ms); Sangam/GeezaPro unchanged.

Final PR standings (clang release rig, benchmark-shape wall vs hb ot — day-start figures in parens): **GeezaPro 98.2% (85.5), Devanagari Sangam 96.4% (86.3), Menlo 98.3% (90), Lucida Grande 104% — ahead of harfbuzz (98.7)**.

Identity: byte-identical on all four AAT fonts + PingFang/CJK; 6,267 tests; clippy clean; no_std+libm builds.

*Written by Claude (via Behdad's account).*

behdad:
Closing per Behdad's review: the decoded transition tables and dense glyph sets buy their speed with per-face/per-buffer memory (up to ~430KB/face for large morx machines), which is not an acceptable tradeoff for this library. The measurement findings survive and are being re-cut as three follow-ups: (1) the safe-to-break cascade factors per-state/per-class into O(rows+cols) tables (~hundreds of bytes), approved for both HarfRust and HarfBuzz; (2) entry() parse slimming in read-fonts (zero memory); (3) U32Set insert-path optimization in read-fonts (zero memory — inline hints alone measured −3.3% instructions on GeezaPro). *Written by Claude (via Behdad's account).*

PR #441, opened 2026-08-23
[aat] Factor the safe-to-break probes into per-class/per-state tables

behdad:
The safe-to-break conditions in the AAT drive loops re-fetch up to two extra state-table entries per transition. Both probes factor by a single axis — condition 2c's "wouldbe" entry depends only on the *class*, condition 3's end-of-text entry only on the *state* — so this builds two small per-machine tables once per face and replaces the probes with lookups:

- **wouldbe**: one `u32` per class packing `new_state | advance << 16` (sentinel when absent/actionable) — the whole 2c/2c′/2c″ check becomes a single compare;
- **eot_tail**: one bit per state ≥ 64, extending the existing `start_end_safe_to_break` word — **empty for machines with ≤ 64 states**, which is most of them.

Storage is O(classes) + O(states): typically tens to hundreds of bytes per machine (Lucida's 55-class/741-state contextual machine: ~330 bytes). Applied to morx, kerx, and legacy kern drive loops; class clamping and every `entry()` error convention preserved exactly (including the asymmetric Err conventions between the <64 bitmask and the ≥64 probe).

Measurements (clang release rig, instructions, single-process): **Devanagari Sangam −2.5%** — its 84-class contextual machine exercises the probes on most transitions; GeezaPro/Menlo/Lucida within this rig's ±0.5% code-layout noise band. Wall-clock deferred to your rig (mine currently has background builds polluting timings). Byte-identical outputs on GeezaPro, Devanagari Sangam, Menlo, Lucida Grande, and PingFang/CJK; 6,267 tests; clippy clean; no_std builds.

An hb twin of this change is being prepared next, per discussion.

### DIFF
--- a/harfrust/src/hb/aat/layout_common.rs
+++ b/harfrust/src/hb/aat/layout_common.rs
@@ -52,7 +52,7 @@ pub struct AatApplyContext<'a> {
     pub(crate) second_set: Option<&'a U32Set>,
     pub(crate) machine_class_cache: Option<&'a ClassCache>,
     pub(crate) start_end_safe_to_break: u64,
-    pub(crate) safe_to_break_cache: Option<&'a SafeToBreakCache>,
+    pub(crate) safe_to_break_accel: Option<&'a SafeToBreakAccel>,
 }
 
 impl<'a> AatApplyContext<'a> {
@@ -76,7 +76,7 @@ impl<'a> AatApplyContext<'a> {
             second_set: None,
             machine_class_cache: None,
             start_end_safe_to_break: 0,
-            safe_to_break_cache: None,
+            safe_to_break_accel: None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


dfrg:
LGTM but I wonder if it’s worth going SoA here and packing the data for all subtables into two (or three, if we store the n_classes and start indices in a Vec rather than directly in the subtable cache) Vecs in a single cache on the top level table. Should reduce the 56 byte per-table overhead down to 12 or so. We might also be able to remove the unwraps on every access, saving a branch. Let me know what Claude says about this :)

behdad:
Implemented SoA but it's measurably slower. Investigating, but will probably revert and merge.

dfrg:
> Implemented SoA but it's measurably slower. Investigating, but will probably revert and merge.

I'll defer to you. Either way works for me.

behdad:
> > Implemented SoA but it's measurably slower. Investigating, but will probably revert and merge.
> 
> I'll defer to you. Either way works for me.

Working on it. It was being hooked in the wrong place to begin with.

PR #443, opened 2026-08-23
[api] Expose Shaper::from_font_instance

behdad:
Lets callers that shape repeatedly build the `Shaper` — the per-call `OtTables`/`AatTables` view — once and reuse it, instead of reconstructing it inside `shape()` on every call. Experimental, gated like `shape()`.

Motivation: in the HarfBuzz bridge, the view reconstruction is ~800 instructions of dependent-load-heavy table parsing per shape call. Caching the shaper keyed on hb's variation-coords serial measures (clang release rig, harfrust shaper): **Roboto/en-words −6.2% instructions / −10.7% cycles**, Roboto/thelittleprince −1.0%/−4.2%, Devanagari Sangam/hi-words −2.4%/−4.4%, Menlo −1.4%/−2.8%, Lucida −1.1%/−2.1%, GeezaPro −0.4%/−0.6% — short-buffer workloads gain most. Outputs byte-identical throughout (including Roboto cross-engine vs hb-ot). The hb-side consumer PR follows once this is in a release.

### DIFF
--- a/harfrust/src/hb/face.rs
+++ b/harfrust/src/hb/face.rs
@@ -498,6 +498,15 @@ pub struct hb_font_t<'a> {
 }
 
 impl<'a> crate::Shaper<'a> {
+    /// Builds a shaper for the font instance, reusing the instance's
+    /// cached shaping data. The shaper borrows the instance; callers
+    /// that shape repeatedly can build it once and reuse it across
+    /// calls.
+    #[cfg(feature = "experimental_font_api")]
+    pub fn from_font_instance(font: &'a crate::font::FontInstance) -> Option<Self> {
+        Self::from_font(font)
+    }
+
     pub(crate) fn from_font(font: &'a crate::font::FontInstance) -> Option<Self> {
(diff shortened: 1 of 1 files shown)


behdad:
@dfrg Does this make sense to you? Allows us to retain more stuff in the hb-hr bridge.

dfrg:
I like the idea but I'm a bit leery of tying the new and old APIs together without more thought. This would also require all of HB, HR and fontations to roll together in Chrome. If we're doing a breaking change in the hb-hr layer, then I'd like to add some more value-- for example, moving to a single `Buffer` type. Let's hold this one for a few days?

In the meantime, we might be able to reduce the cost of a shape call by memoizing the blocklist checks in `OtCache` and `AatCache` rather than computing them every time we materialize `OtTables` and `AatTables`. That might shave a few cycles without breaking anything.

behdad:
Humm okay. I'll send it back.

The https://github.com/harfbuzz/harfrust/pull/441 SoA was a regression but when moving it on top of this it became a win.

PR #445, opened 2026-08-23
[aat] Factor safe-to-break probes into the face cache

behdad:
The AAT drive loops currently re-fetch up to two state-table entries per transition. Both probes factor by one axis, so this precomputes their results once per face:

- `wouldbe`: one packed `u16` per class for condition 2c.
- `eot_tail`: one bit per state at or above 64 for condition 3.

The storage is entirely face-level and independent of the experimental `Shaper` API:

- `AatCache` owns two packed vectors containing data for every morx, kern, and kerx state-machine subtable.
- Each existing top-level subtable cache contains its own 16-byte descriptor with explicit start/end pairs for both vectors.
- The apply loops resolve the slice view before the glyph-intersection scan, keeping the range loads off the immediate drive-loop dependency chain.
- The transition loops retain the packed comparison and no longer unwrap an optional acceleration object on each access.

## Previous attempts

This supersedes both closed attempts:

- #441 put 12-byte descriptors in a third centralized vector. Selecting every subtable required another indexed metadata lookup, reading the following descriptor, and validating two ranges on every shape call.
- #444 pre-resolved 40-byte slice views in `AatTables`. That removed repeated setup when combined with #443 and harfbuzz/harfbuzz#6199, but depended on the experimental cached-`Shaper` work and duplicated the views per constructed Shaper.

The first revision of this PR used a packed 12-byte descriptor. Although it retired fewer instructions, unpacking its lengths immediately before the drive loop reduced IPC enough to lose in elapsed-time measurements. The current 16-byte descriptor stores both range ends explicitly and resolves the view before scanning the buffer glyph set.

## Comparison with #441

Using the clang release HarfBuzz `main` build and `hb-shape -O "" -n 100 --shaper harfrust` (five-run `perf stat` averages), current #445 versus the final centralized-SoA commit from #441:

| Workload | instructions | cycles |
|---|---:|---:|
| Menlo / en-thelittleprince | -0.52% | -5.43% |
| Lucida Grande / en-thelittleprince | -0.61% | -3.35% |
| Geeza Pro / fa-thelittleprince | -0.29% | -0.59% |
| Devanagari Sangam / hi-words | -0.87% | -1.26% |

Cycle measurements are non-interleaved local results and should be confirmed independently. Output hashes match #441 for all four workloads.

The face cache uses 16 bytes of descriptor storage and two shared allocations instead of 56 bytes plus two allocations per subtable.

Validation:

- `cargo test` (6,268 tests)
- `cargo clippy --all-features --all-targets -- -D warnings`
- `cargo build --no-default-features --features=libm`

### DIFF
--- a/harfrust/src/hb/aat/layout_common.rs
+++ b/harfrust/src/hb/aat/layout_common.rs
@@ -52,7 +52,7 @@ pub struct AatApplyContext<'a> {
     pub(crate) second_set: Option<&'a U32Set>,
     pub(crate) machine_class_cache: Option<&'a ClassCache>,
     pub(crate) start_end_safe_to_break: u64,
-    pub(crate) safe_to_break_cache: Option<&'a SafeToBreakCache>,
+    pub(crate) safe_to_break_accel: Option<&'a SafeToBreakAccel>,
 }
 
 impl<'a> AatApplyContext<'a> {
@@ -76,7 +76,7 @@ impl<'a> AatApplyContext<'a> {
             second_set: None,
             machine_class_cache: None,
             start_end_safe_to_break: 0,
-            safe_to_break_cache: None,
+            safe_to_break_accel: None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


dfrg:
I like the shape of this one. LGTM

PR #448, opened 2026-08-24
[shape] Add an in-place reusable buffer API

behdad:
Adds shape_in_place() under experimental_font_api while preserving the existing shape() API.

The new API borrows a UnicodeBuffer, shapes its allocation-bearing state in place, and returns a scoped GlyphBufferRef. The guard exposes the result slices and clears the buffer on drop, so the caller retains the UnicodeBuffer type-state and can immediately reuse its allocations.

This is intended for reusable FFI caches such as the HarfBuzz bridge. It avoids moving the 304-byte internal buffer out of and back into a Box on every shape.

With the corresponding HarfBuzz bridge change, measured against the same cached-buffer baseline:

- Roboto prince: -0.33% instructions, -0.51% cycles
- Roboto words: -2.19% instructions, -3.94% cycles
- all eleven AAT/OT benchmark pairs improved or were neutral
- callgrind on Roboto words: about 301 fewer instructions per shape, including about 65 fewer memcpy instructions

Validation:

- cargo test -p harfrust --all-features (6,271 tests)
- cargo clippy -p harfrust --all-features --all-targets -- -D warnings
- HarfBuzz HarfRust API tests (4 subtests)
- output parity with the OT shaper on all eleven benchmark inputs

### DIFF
--- a/harfrust/src/hb/buffer.rs
+++ b/harfrust/src/hb/buffer.rs
@@ -2234,6 +2234,49 @@ impl GlyphBuffer {
     }
 }
 
+/// A scoped view of shaping results in a reusable [`UnicodeBuffer`].
+///
+/// The buffer is cleared automatically when this view is dropped, preserving
+/// the `UnicodeBuffer` type-state while allowing its allocation-bearing state
+/// to remain in place during shaping.
+#[cfg(feature = "experimental_font_api")]
+pub struct GlyphBufferRef<'a>(pub(crate) &'a mut hb_buffer_t);
+
+#[cfg(feature = "experimental_font_api")]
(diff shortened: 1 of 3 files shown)


behdad:
Companion HarfBuzz bridge PR: harfbuzz/harfbuzz#6202.

behdad:
Closing because UnicodeBuffer and GlyphBuffer are expected to be unified soon; adding a temporary borrowed typestate API would create unnecessary churn.

behdad:
It's interesting how much this seems to weigh in the Roboto/words bench. Closing since we intend to land https://github.com/harfbuzz/harfrust/pull/298 eventually.

PR #451, opened 2026-08-24
[aat] Add mort shaping support

behdad:
## Summary

- cache and apply legacy mort tables in the AAT shaper
- share chain selection and state-machine semantics with morx through small common traits
- retain format-specific adapters for mort absolute offsets and compact entry indexes
- prefer morx when a font contains both tables, matching HarfBuzz
- add paired mort and morx shaping coverage for all five subtable types

## Dependencies

- Fontations mort parser: https://github.com/googlefonts/fontations/pull/2074
- upstream test fixtures and converter: https://github.com/unicode-org/text-rendering-tests/pull/108
- HarfBuzz fixture import: https://github.com/harfbuzz/harfbuzz/pull/6206

The final dependency commit pins the exact Fontations PR revision so CI can exercise the implementation now. It should be switched back to a crates.io read-fonts release before merge.

## Testing

- cargo check --workspace --all-features --all-targets
- cargo test --workspace --all-features
- cargo clippy --workspace --all-features --all-targets -- -D warnings
- no-std libm build


dfrg (approved):
This looks good, sharing as much as possible with morx.

PR #452, opened 2026-08-24
[kerx] Handle tuple kerning

behdad:
Match HarfBuzz handling of `kerx` subtables with tuple kerning values.

HarfBuzz implements this "without the variation part": tuple-backed values are offsets to an array and shaping applies the first value. This stops skipping those subtables, resolves tuple offsets for formats 0 and 2, and lets the existing format 1 and 6 tuple handling run.

The tuple resolver validates the complete tuple array before reading its first value.

Tests:
- `cargo test -p harfrust`
- `cargo clippy -p harfrust --all-targets -- -D warnings`

### DIFF
--- a/harfrust/src/hb/aat/layout_kerx_table.rs
+++ b/harfrust/src/hb/aat/layout_kerx_table.rs
@@ -160,9 +160,7 @@ fn resolve_tuple_kerning(value: i32, tuple_count: u32, data: FontData<'_>) -> Op
     }
 
     let offset = usize::try_from(value).ok()?;
-    let byte_len = usize::try_from(tuple_count).ok()?.checked_mul(2)?;
-    let tuple = data.slice(offset..offset.checked_add(byte_len)?)?;
-    tuple.read_at::<i16>(0).ok().map(i32::from)
+    data.read_at::<i16>(offset).ok().map(i32::from)
 }
 
 impl SimpleKerning for Subtable0<'_> {
@@ -823,6 +821,6 @@ mod tests {
 
         assert_eq!(value, 22);
         assert_eq!(subtable.resolve_tuple_kerning(value, 2), Some(-100));
-        assert_eq!(subtable.resolve_tuple_kerning(value, 3), None);
+        assert_eq!(subtable.resolve_tuple_kerning(value, 3), Some(-100));
     }
 }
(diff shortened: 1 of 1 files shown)


harfrust/src/hb/aat/layout_kerx_table.rs
```diff
@@ -150,13 +146,32 @@ pub(crate) fn apply(c: &mut AatApplyContext) -> Option<()> {
...
+// are offsets to an array of tupleCount FWORDs, of which it applies the first.
+fn resolve_tuple_kerning(value: i32, tuple_count: u32, data: FontData<'_>) -> Option<i32> {
+    if tuple_count == 0 {
+        return Some(value);
+    }
+
+    let offset = usize::try_from(value).ok()?;
+    let byte_len = usize::try_from(tuple_count).ok()?.checked_mul(2)?;
+    let tuple = data.slice(offset..offset.checked_add(byte_len)?)?;
+    tuple.read_at::<i16>(0).ok().map(i32::from)
```
dfrg: Seems like `byte_len` and `tuple` are used just to validate the full array? Any reason why we don't just read at the offset?

dfrg (approved):
LGTM, with a question.

harfrust/src/hb/aat/layout_kerx_table.rs
behdad: Codex first: Yes—that is what those two variables are doing. I added the full-array check to mirror HarfBuzz's `kerxTupleKern()`, which calls `sanitizer.check_array(pv, tupleCount)` before using the first `FWORD`. A direct read at the offset would still be memory-safe, but it would accept a truncated tuple vector that HarfBuzz rejects (the `tuple_count = 3` assertion below exercises that distinction). I was aiming to preserve HarfBuzz's malformed-font behavior here; if matching read-fonts' usual lazy-validation style is preferable, I’m also happy to simplify it.

behdad: I'll instruct it to simplify.

behdad: Simplified in 9a85b894: this now reads the first `FWORD` directly at the offset and leaves the remaining tuple coordinates unvalidated, consistent with read-fonts' lazy-validation style. I also updated the test accordingly.

PR #454, opened 2026-08-24
[arabic] Synthesize fallback GSUB lookups

behdad:
## Summary

- synthesize the Arabic `init`, `medi`, `fina`, `isol`, and `rlig` fallback lookups directly as OpenType SingleSubst and LigatureSubst byte streams, cached on the shape plan
- avoid the `write-fonts` dependency entirely
- support the legacy Simplified and Traditional Arabic symbol cmaps through the HarfBuzz PUA compatibility maps
- generate the fallback and PUA data through the HarfBuzz `--rust` generators added in harfbuzz/harfbuzz#6207
- enable all 13 `arabic-fallback-shaping.tests` cases from the shared HarfBuzz shaping suite
- support fallback synthesis in `no_std` builds too, cached exactly once per shape plan with `spin::Once`

This implements the portable Unicode presentation-form path. It intentionally does not include HarfBuzz's Windows-only hard-coded Win1256 GSUB blob.

## Testing

- `cargo test --workspace --all-features`
- `cargo check -p harfrust --no-default-features --features libm`
- `cargo test -p harfrust --no-default-features --features libm --test shaping arabic_fallback_shaping`
- `cargo test -p harfrust --no-default-features --features libm maps_legacy_arabic`
- `cargo clippy --workspace --all-features -- -D warnings -A clippy::multiple-crate-versions`
- `cargo fmt --check`


harfrust/src/hb/charmap.rs
```diff
@@ -116,6 +101,44 @@ impl<'a> Charmap<'a> {
...
+        super::ot_shaper_arabic_pua::_hb_arabic_pua_trad_map(c)
+    };
+    u32::from(mapped)
+}
+
+fn map_subtable(subtable: &CmapSubtable<'_>, c: u32) -> Option<GlyphId> {
+    match subtable {
+        CmapSubtable::Format0(table) => table.map_codepoint(c),
+        CmapSubtable::Format6(table) => table.map_codepoint(c),
+        CmapSubtable::Format10(table) => {
```
dfrg: I think this one can just call `table.map_codepoint(c)` now. It was added to `read-fonts` at some point.

khaledhosny (approved):

harfrust/Cargo.toml
```diff
@@ -22,7 +23,7 @@ icu_normalizer = { version = "2.2.0", optional = true }
 
 [features]
 default = ["std"]
-std = ["read-fonts/std"]
+std = ["read-fonts/std", "dep:write-fonts"]
```
dfrg: This should gate on it's own feature, off by default. The `write-fonts` dependency is fairly heavy and most users probably don't want it.

behdad:
I think I'll ask agents to rewrite this without write-fonts dep.

harfrust/src/hb/ot_shaper_arabic_fallback.rs
```diff
@@ -0,0 +1,170 @@
...
+    ot_layout_gsubgpos::OT::hb_ot_apply_context_t,
+    ot_shape_plan::hb_ot_shape_plan_t,
+    ot_shaper_arabic_table::{
+        LIGATURE_3_TABLE, LIGATURE_MARK_TABLE, LIGATURE_TABLE, SHAPING_TABLE,
+    },
+};
+use write_fonts::{
+    dump_table,
+    tables::{
+        gsub::builders::{LigatureSubBuilder, SingleSubBuilder},
```
dfrg: These are both fairly simple tables. Could we possibly have some internal structure for these that doesn't have to roundtrip through OpenType so we can avoid the write-fonts dep? Not necessarily for this PR but seems worth looking into.

dfrg (approved):
Nice work! Generally LGTM assuming the whole thing is gated on a anew off-by-default feature and not on std.

harfrust/src/hb/charmap.rs
behdad: Done in c304350: Charmap now calls CmapSubtable::map_codepoint directly.

harfrust/Cargo.toml
behdad: Done in c304350: removed write-fonts entirely, so no new opt-in feature is needed. The fallback retains its existing std gate.

harfrust/src/hb/ot_shaper_arabic_fallback.rs
behdad: Done in c304350: the fallback now directly serializes the SingleSubst and LigatureSubst lookup byte streams. It has serializer parsing tests plus the existing shaping coverage.

behdad:
@dfrg PTAL. No write-fonts.

Cargo.toml
```diff
@@ -23,6 +23,7 @@ inherits = "release"
 
 [workspace.dependencies]
 read-fonts = { version = "0.43.2", default-features = false, features = ["experimental_font_api"] }
+spin = { version = "0.12.3", default-features = false, features = ["once"] }
```
dfrg: The spin crate is not available in Chrome third_party/rust which is why read-fonts uses once_cell and the racing lock. See https://github.com/googlefonts/fontations/blob/main/read-fonts/src/model/once.rs

dfrg: Just a note to our future selves: having any sort of no_std once allows us to make our no_std lookup cache lazy as well. It currently caches all lookups at creation.

dfrg (approved):
Approving but I still don't think we should gate this on `std`. Let's add a new `arabic_fallback` feature for this. Without the write-fonts dependency, I don't have a preference as to whether the feature is enabled by default-- up to you.

We could also make this a runtime option on `ShapeOptions`? Not sure if that makes sense.

behdad:
> Approving but I still don't think we should gate this on `std`.

It's not. There was a stale comment.

> Let's add a new arabic_fallback feature for this. Without the write-fonts dependency, I don't have a preference as to whether the feature is enabled by default-- up to you.

I highly recommend against it. If we were to add one for this, there's like dozens of other options we should add of the same caliber.

> We could also make this a runtime option on ShapeOptions? Not sure if that makes sense

No please.

dfrg:
LGTM! 🚢

PR #458, opened 2026-08-27
Remove private codes from `Script::from_iso15924_tag`

robertbastian:
These private codes should not be used in interchange, which is what that constructor is.

### DIFF
--- a/harfrust/src/hb/common.rs
+++ b/harfrust/src/hb/common.rs
@@ -356,12 +356,6 @@ impl Script {
         }
 
         Some(match &tag.to_be_bytes() {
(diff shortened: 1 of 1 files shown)


behdad:
What's the rationale? I know these are dated, but if removing we should also remove in HarfBuzz.

robertbastian:
`Qaai` has not been returned by ICU since Unicode 5.2 (2009), `Qaac` not since at least Unicode 4.1 (2005). At some point these ICU-private tags should be removed from downstream code.

behdad:
Thanks. Fixed in HarfBuzz as well.

PR #459, opened 2026-09-01
Add a unified Buffer type

dfrg:
Apologies for the churn. The only files actually modified are buffer.rs, face.rs and the new buffer tests. The remainder are just fallout from renaming the buffer type for public exposure.

Renames the internal `hb_buffer_t` to `Buffer` and exposes it, matching HarfBuzz's `hb_buffer_t`. Every internal method and field drops to `pub(crate)`, and the public surface is the set of operations HarfBuzz's hb-buffer.h offers.

`Buffer` carries a `BufferContentType` recording whether it holds input characters or the glyphs produced by shaping. Shaping sets it to `Glyphs`, and shaping a buffer that already holds glyphs is a no-op. This makes the buffer usable across the whole shaping cycle from a single type, which the typed pair cannot express: notably, glyph infos are now readable before shaping, and the segment properties remain readable after it.

`UnicodeBuffer` and `GlyphBuffer` are unchanged. They convert into `Buffer` infallibly, and back out through `TryFrom`, which fails with `WrongContentType` when the buffer holds the other kind of content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This development version matches HarfBuzz [v14.3.1](https://github.com/harfbuzz/
   `BufferContentType` describing whether it holds input characters or shaped
   glyphs, and is shaped in place with `Buffer::shape`. `UnicodeBuffer` and
   `GlyphBuffer` are unchanged, and convert to and from `Buffer`.
+- `Buffer::shape` returns `Result<(), ShapeError>`, reporting a buffer that
+  already holds glyphs, a buffer with no direction, a font with nothing to
+  shape with, a plan built for other properties, and shaping that ran past its
+  length, operation or nesting limits.
 
 ## [0.13.3] - 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 This development version matches HarfBuzz [v14.3.1](https://github.com/harfbuzz/harfbuzz/releases/tag/14.3.1).
 
+- Add a unified `Buffer` type, matching HarfBuzz's `hb_buffer_t`. It carries a
+  `BufferContentType` describing whether it holds input characters or shaped
+  glyphs, and is shaped in place with `Buffer::shape`. `UnicodeBuffer` and
+  `GlyphBuffer` are unchanged, and convert to and from `Buffer`.
+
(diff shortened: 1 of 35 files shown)


behdad:
Looks good to me!

> shaping a buffer that already holds glyphs is a no-op.

I don't like this. I think it should return `WrongContentType`.

dfrg:
> I don't like this. I think it should return `WrongContentType`.

Sounds good. If we’re going to return errors from `shape` then we might as well capture them all. Will update.

behdad:
> > I don't like this. I think it should return `WrongContentType`.
> 
> Sounds good. If we’re going to return errors from `shape` then we might as well capture them all. Will update.

Ooh if `shape` is currently infallible, fine with me to keep it that way!

dfrg:
> > > I don't like this. I think it should return `WrongContentType`.
> > 
> > 
> > Sounds good. If we’re going to return errors from `shape` then we might as well capture them all. Will update.
> 
> Ooh if `shape` is currently infallible, fine with me to keep it that way!

Pushed a commit with `shape` returning `Result<(), ShapeError>`. PTAL, I'll revert if you'd rather go that way.

behdad:
> > > > I don't like this. I think it should return `WrongContentType`.
> > > 
> > > 
> > > Sounds good. If we’re going to return errors from `shape` then we might as well capture them all. Will update.
> > 
> > 
> > Ooh if `shape` is currently infallible, fine with me to keep it that way!
> 
> Pushed a commit with `shape` returning `Result<(), ShapeError>`. PTAL, I'll revert if you'd rather go that way.

Thanks. Just to make sure, `LimitsExceeded` still lives the buffer shaped, right? Because not producing output is not an acceptable result to me. I personally prefer not exposing that one but that's me.

dfrg:
> Thanks. Just to make sure, LimitsExceeded still lives the buffer shaped, right? Because not producing output is not an acceptable result to me. I personally prefer not exposing that one but that's me.

Right, the buffer content is left untouched in that case. Whatever was shaped up until we ran out of budget. I'll remove that variant as an error.

dfrg:
Done

behdad:
LGTM.

So what's the plan? To retire UnicodeBuffer / GlyphBuffer at some point?

dfrg:
> So what's the plan? To retire UnicodeBuffer / GlyphBuffer at some point?

Yeah, we’ll mark them deprecated for a release or two and then delete them. HB should probably wait for one HR roll in Chrome before adapting.

Issue #412, opened 2026-08-08
Sync with HarfBuzz 14.3.0


khaledhosny:
There are at least to shaping-affecting changes in HarfBuzz 14.3.0. Worse reviewing all the changes since last sync and porting over anything relevant to HarfRust.

behdad:
Right. Typically it's just a LLM job. I'll take a look.

Issue #409, opened 2026-07-26
`set_not_found_variation_selector_glyph` produces a `glyph_id` above `u16::MAX`


DRMacIver:
```rust
use harfrust::font::{Font, FontInstance};
use harfrust::{shape, SerializeFlags, ShapeOptions, UnicodeBuffer};

fn main() {
    // small published font with no UVS entry for A + U+FE00
    let font = Font::new(font_test_data::CMAP14_FONT1.to_vec(), 0).unwrap();
    let instance = FontInstance::builder(&font).build();

    let mut buffer = UnicodeBuffer::new();
    buffer.push_str("A\u{FE00}");
    buffer.guess_segment_properties();
    buffer.set_not_found_variation_selector_glyph(0x1_0000);

    let glyphs = shape(&instance, buffer, ShapeOptions::new());
    for info in glyphs.glyph_infos() {
        println!("glyph_id = {}", info.glyph_id);
    }
    println!("serialized = {}", glyphs.serialize(&instance, SerializeFlags::empty()));
}
```

with `harfrust = { version = "0.12", features = ["experimental_font_api"] }` and `font-test-data = "0.8"`. It prints (debug and release alike):

```
glyph_id = 0
glyph_id = 65536
serialized = [gid0=0+1500|gid65536=0+0]
```

`GlyphInfo::glyph_id` is documented as "Guarantee to be <= `u16::MAX`". Here `set_not_found_variation_selector_glyph(0x1_0000)`, when the variation selector (`A` + `U+FE00`) isn't a registered sequence in the font, gives a `glyph_id` of 65536 — above the documented maximum. The value surfaces in `glyph_infos()` and in the serialized output.

(Originally reported to rustybuzz as harfbuzz/rustybuzz#168, where the maintainer pointed here.)

Tested on harfrust 0.12.0.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

behdad:
> "Guarantee to be <= u16::MAX".

We should kill any such suggestions.

> BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

I think that would be nice. Yes. Thanks.

PR #461, opened 2026-09-02
Add hegel property tests

DRMacIver:
Following up on #409, where you said tests for that bug would be nice to have. Claude Code wrote the tests and this description. Please read this as an offer of code rather than an imposition: if it is more than you want to own, or the wrong shape for the repo, say so and no offence will be taken.

The last commit adds a differential harness that shapes the same input through HarfBuzz itself, which brings in a dev-dependency on `harfbuzz_rs`. It is `#[ignore]`d, because `harfbuzz_rs` vendors HarfBuzz 11.3.3 while HarfRust tracks something newer (#412), so a divergence it reports has to be triaged against current HarfBuzz before it means anything. Say if you'd rather it went.

Six tests are `#[ignore]`d: that harness, a pin for #409, and four things I haven't filed. The #409 pin is the not-found variation selector glyph above `u16::MAX` reaching `GlyphInfo::glyph_id`, and it pins the disagreement between the code and the docs rather than the bound you said should go. Each of the four unfiled ones has its reason in a comment above it. One is a cmap entry resolving to glyph 0, which HarfRust treats as a hit and HarfBuzz treats as a miss, and which may belong in `read-fonts` rather than here. One is the `i64` multiply in `Scale::scale_by_mult`, which is unguarded but not reachable through the public API at any advance that fits in a `u16`. The other two are cases where breaking at a cluster that isn't flagged `UNSAFE_TO_BREAK` still changes the result, which HarfBuzz 11.3.3 doesn't uphold either. Say if you'd like any of those filed as issues, or the pins dropped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@
 /Cargo.lock
 benches/Cargo.lock
 /.cargo
+# hegeltest's saved failing inputs
 .hegel/
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /Cargo.lock
 benches/Cargo.lock
 /.cargo
+.hegel/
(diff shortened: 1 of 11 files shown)


DRMacIver:
Apologies, on further review I've found a bunch of issues with this PR. I'm closing it for now and will open a fresh one when I'm more satisfied with its quality.

PR #462, opened 2026-09-03
Optimize repeated hr-shape runs

behdad:
## Summary

- cache shape plans by direction, script, language, and features
- reuse Unicode buffer allocations across iterations and input lines
- support `-O ""` to suppress glyph serialization, matching `hb-shape`

## Performance

Roboto Regular, `en-words.txt`, 100 iterations:

- before: 2.76s
- after: 0.88s
- `hb-shape`: 0.81s

This reduces the gap from roughly 3.4x to 1.1x.

## Verification

- `cargo test -p hr-shape`
- `cargo clippy -p hr-shape --all-targets --all-features -- -D warnings`
- release build and original command-line benchmark
- verified `-O ""` output against `hb-shape`


dfrg (approved):
LGTM

PR #463, opened 2026-09-04
[DO-NOT-MERGE] HR-ng: Crazy rework of lookup caching and scan loop

dfrg:
Quantification of the classic time vs space trade off: use more memory to buy performance. This implements a lookup/subtable compiler that transforms coverage and class defs into optimal in memory representations, interns them and then chooses optimal routing during the scan loop. Heap cost is 2x current HR on average (0.9x for Devanagari!). Performance to be measured.


behdad:
This is insane, in the best of the ways! I'll review soon. In the mean time, some beauty:

image

https://docs.google.com/spreadsheets/d/1bROQNTq4SNv3WhASdlJcffoBsTLuWlpr4b_smWShc4A/edit

behdad:
I am Codex, an OpenAI coding agent, and I prepared the following independent review report.

# `perf/compile-more` review and HarfBuzz assessment

Scope: `perf/compile-more` at `c992fb01`, compared with merge-base
`b4392db6` on `origin/main` (35 commits; 24 files; +7,998/-71 lines). The
HarfBuzz assessment started from local HarfBuzz `main` at `80540ed4`
(2026-09-03); its non-lookup status is updated through merged PR #6231.

## Executive assessment

The branch is a substantial and credible shaping optimization. It does not JIT
machine code. It lazily *partially evaluates* each used GSUB/GPOS lookup into
typed, heap-owned metadata: adaptive coverage/class representations, exact
reachability sets, rejection summaries, direct apply function pointers, and
optional subtable dispatch. Large cold payloads remain in the font and are read
only after the metadata says a match is plausible.

The design is worth landing in HarfRust, but it should not yet be enabled by
default. Two correctness bugs must be fixed first: probabilistic interner
equality can mis-shape on a hash collision, and a failed nested lookup
compilation has no interpreted fallback. There is also an intentional change to
`UNSAFE_TO_CONCAT` flags that needs an explicit compatibility decision. After
those are addressed, land the lookup compiler behind an opt-in feature, retain
the interpreter as fallback/oracle, and default-enable only after fuzzing and
cold-start/memory measurements.

For HarfBuzz, port the data-structure and rejection ideas incrementally rather
than transplanting the Rust compiled engine. HarfBuzz already has a lazy
per-lookup accelerator, typed subtable apply functions, digests, and several
external subtable caches. The highest-value additions are the next-glyph
lookup digest, contextual rule-set summaries, PairPos summaries, and adaptive
exact coverage/class representations. The four non-lookup pass fusions should
be evaluated separately. The two applicable loop fusions have since landed in
HarfBuzz through [PR #6231](https://github.com/harfbuzz/harfbuzz/pull/6231).

## Part I — What the HarfRust branch does

### 1. GSUB/GPOS lookup compilation

The new path is under `harfrust/src/hb/ot/compile/`; `OtCache` owns one
`Program` for GSUB and one for GPOS. Each program has one
`OnceLock<Option<Box<CompiledLookup>>>` per lookup. The first use compiles the
lookup; subsequent shapes are lock-free cache hits. A mutex protects compiler
scratch space only on a miss. Scratch vectors are reused across the lookups of
one table pass, then released.

The compiled form is lifetime-free: it stores table-relative offsets and owned
indexes, then borrows the current font bytes while applying. This is a useful
split:

- Compile hot control data: lookup properties, coverage membership/rank,
  class maps, reachability, rule summaries, format selection, and dispatch.
- Leave bulky/cold payloads in the font: substitute arrays, ligature records,
  value records, anchors, and most contextual rules.
- Reuse existing shaping primitives for matching, ligation, value application,
  attachment, recursion budgets, and buffer mutation.

`CompiledLookup` contains the lookup properties and direction, a union reach
set and three-word reach digest, a three-word “possible next glyph” digest,
compiled subtables, and optional glyph-to-subtable dispatch. Each subtable has
an interned coverage, a format-specific apply function pointer, a specialized
coverage gate, typed metadata, and sometimes an exact set of possible second
inputs.

#### Adaptive indexes

Three structures replace repeated parsing and binary searches:

- `GlyphSet` (membership only): empty, contiguous range, bitmap, ranges, or
  sorted glyphs.
- `Coverage` (membership plus coverage index): empty, contiguous range, ranked
  bitmap, ranges, or sorted glyphs.
- `ClassMap`: empty, dense 8-bit, dense 16-bit, or ranges.

The compiler chooses a fast form whose size is at most
`max(384 bytes, 8 × smallest representation)`; reach-set scans have a 4,224-byte
budget. Keeping membership-only sets separate from ranked coverages matters:
scanning a ranked coverage instead reportedly cost 14% in one experiment.

GSUB and GPOS share an `Arc<Interner>` for these structures. Interning removes
duplicate decoded tables, although measured cross-table sharing was only 0–3%
of distinct tables.

#### Hot lookup driver

Before scanning, the driver rejects a lookup if either of these cannot
intersect the buffer digest:

1. any glyph that may start the lookup; or
2. any glyph that may occur in the lookup's second constrained position.

The second test (`pair_digest`) is the strongest isolated optimization recorded
in the branch: its incremental elapsed-time mean is 0.848 (about 15%). It is
soundly disabled when any subtable does not constrain a second input, for
reverse lookups, and for one-component ligatures.

For a surviving lookup, the scanner hoists the chosen reach representation,
face, mask, and properties out of the glyph loop. It tests exact reach, mask,
and glyph properties before calling a subtable. A one-subtable lookup avoids a
redundant coverage check when the scan has already established membership. With
at least four selective subtables, a bounded optional dispatch table maps an
input glyph to its ordered candidate subtable row (maximum 4,096 pairs).

The subtable format and coverage representation are selected once at compile
time through function pointers. This removes repeated enum construction,
format branching, offset chasing, and generic coverage lookup from the hot
loop. It is “JIT-like” specialization, but emits no executable code.

#### Contextual and pair-position summaries

- Context/chained rule sets get a one-word summary of values accepted at the
  first tested position. A miss skips the entire rule set. This accounts for a
  further 4–9 percentage points on the Nastaliq cases in the commit history.
- PairPos format 1 gets a set summary for the second glyph before binary
  searching a pair set.
- PairPos format 2 gets a per-class1-row summary of class2 values whose raw
  value records are nonzero. This avoids reading mostly-zero matrix rows while
  preserving the fact that a zero record may still count as a match.

Nested contextual application remains in the compiled engine and reuses the
host operation/nesting budgets and matching code.

#### Format coverage

All 13 logical lookup families compile successfully, but not all receive a
native compiled apply implementation:

| Table | Treatment |
|---|---|
| GSUB Single | Delta is compiled; replacement arrays remain borrowed. |
| GSUB Multiple/Alternate | Coverage is compiled; sequence/set payload stays borrowed. |
| GSUB Ligature | Coverage and possible second components are compiled; ligature records stay borrowed. |
| GSUB Context/ChainContext | Class/coverage indexes and rule summaries are compiled; shared host rule walkers apply records. |
| GSUB ReverseChain | Input/backtrack/lookahead sets are compiled; substitutes stay borrowed. |
| GPOS Single | Coverage and value metadata are compiled; value data stays borrowed. |
| GPOS Pair | Coverage/class maps and pair/row summaries are compiled; records/matrix stay borrowed. |
| GPOS Cursive | Coverage/offsets are compiled; existing attachment geometry is reused. |
| GPOS Mark-to-* | The outer lookup is accelerated, but application falls through to the existing interpreted mark routine. |
| Extension | Unwrapped during compilation. |

### 2. Other optimizations, independent of GSUB/GPOS compilation

Commit `eabdb6dd` makes four unrelated whole-pass changes:

1. Initialize masks while assigning Unicode properties, removing a separate
   O(n) pass.
2. Populate GDEF glyph properties while rebuilding the buffer digest on the
   non-`morx` path.
3. Avoid zeroing the entire position buffer before horizontal positioning.
   Built-in batch advance code overwrites complete position records; custom
   callbacks are still pre-zeroed, and vertical positioning retains the safe
   initialization path.
4. Remove a redundant glyph-info copy in `map_glyphs_fast`; the input and output
   storage alias in that state.

These should be reviewed and landed separately. The branch reports independent
elapsed-time improvements of 7.8% for Roboto words, 7.0% for Devanagari words,
and 2.3% for Nastaliq words, with bit-exact output across the then-current 6,291
tests.

### 3. Performance and cost

The user-provided benchmark chart shows the new branch below HarfBuzz elapsed
time on all ten plotted font/text pairs (roughly 0.56–0.99× HarfBuzz), whereas
the earlier HarfRust path is roughly 1.07–1.24×. The branch's final committed
seven-case, interleaved comparison against its interpreted path reports a mean
elapsed-time ratio of 0.656: 34.4% less time, equivalently 52.4% more throughput.
Per-case ratios are 0.537–0.800. “40% faster” is therefore directionally fair,
but should always say whether it means elapsed time or throughput.

These are steady-state results. Face/shaper construction and lazy compilation
occur outside or during Criterion warm-up; the timed closure reuses the face,
plan cache, and buffer. The corpus is small (five fonts/seven pairs in the
checked-in harness), uses ten short samples, and does not characterize first
shape, RSS, code size, mobile/WASM, unusual feature sets, directions, or custom
font callbacks. The chart usefully adds Gulzar and Noto Sans Duployan, but raw
methodology is not stored in this branch.

The reported compile cost is approximately 0.2 ms for the Latin case and below
1 ms for Amiri, paid lazily once per face. Representative reached caches are
about 41 KiB (Source Serif) to 268 KiB (Amiri). Four of five cases used about
2–2.9× the comparable interpreted-cache memory; Devanagari was slightly smaller
because its interpreted cache has a large inline variant.

There is also a benchmark configuration bug: the `harfrust` self
dev-dependency enables default features, so Cargo feature unification turns
`compile-path` back on even when invoking `cargo bench ...
--no-default-features --features std`. Fix that before publishing a documented
compiled-vs-interpreted command.

### 4. Correctness and landing review

#### Blockers

1. **Interning uses hashes as equality.** `compile/set.rs` identifies encoded
   tables only by two 64-bit FoldHash values. A collision aliases distinct
   coverages/classes and changes shaping. FoldHash is explicitly a fast,
   non-cryptographic hash; even two hashes are not semantic equality for
   attacker-controlled font bytes. Use the hash only to select a bucket and
   compare exact bytes/canonical decoded keys, or intern by exact table
   identity `(kind, offset, length)` within the face. The latter sacrifices
   little because cross-GSUB/GPOS content sharing is small.

2. **Nested compile failure silently drops an action.** The top-level driver
   correctly falls back to the interpreted lookup when `Program::get()`
   returns `None`. `compile/apply.rs::recurse`, however, chains through
   `program.get(...).and_then(...)`; if a contextual child cannot compile, it
   returns no match instead of calling the host interpreter. Add the same
   fallback before compiled recursion bookkeeping, or require the complete
   reachable lookup closure to compile before using a contextual parent. Add a
   malformed-font regression that exercises this exact path.

3. **Observable buffer-flag difference needs policy.** The exact ligature
   second-glyph filter produces a strict subset of the interpreter's
   `UNSAFE_TO_CONCAT` flags in 7 of 3,169 differential cases; glyph IDs,
   clusters, and positions agree. This may be more precise, but the flags are
   API output. Either preserve current behavior or deliberately change the
   contract with targeted fixtures and release notes. HarfBuzz should require
   exact flag compatibility for an optimization.

#### Cleanup before review

- `RuleFirsts` is built, stored, tested, and memory-accounted but never read by
  shaping. Remove it or implement and benchmark the intended per-rule filter.
- `LengthEffect` scans/stores MultipleSubst sequence lengths but is never used;
  the compiled GSUB path still always uses output/sync. Remove it until an
  in-place path exists.
- compiled MarkTo metadata (`bases`, array offsets, class count) is unused
  because the apply function reparses the original subtable. Retain only the
  outer acceleration or finish the native implementation.
- `Detail::Lean` documentation still claims to disable a pair filter that was
  removed; it now only disables dispatch. The environment-controlled detail
  levels are useful experiments, not a stable library interface.
- Several module comments describe superseded designs (bitmap intersection,
  in-place GSUB, and an unused experimental path). Update them with the code.
- The new `hashbrown`, FoldHash, and `spin` footprint needs explicit `no_std`,
  code-size, and dependency review. Exact-key interning may eliminate the
  riskiest part of this stack.

#### Recommended HarfRust landing sequence

1. Split and land the four non-lookup fusions with their own correctness and
   performance evidence.
2. Fix exact interner equality, nested fallback, benchmark feature selection,
   and the flag-compatibility decision; remove dead metadata and stale knobs.
3. Land the compiler opt-in, keeping the interpreter as the failure fallback
   and differential oracle. Add malformed/adversarial-font, collision-forcing,
   concurrent first-use, all-directions/features, and `no_std` tests.
4. Measure cold first-shape latency, reached and all-lookups memory, code size,
   and representative desktop/mobile/WASM workloads before enabling by default.

The current branch passes `cargo fmt --check`, `git diff --check`, all-feature
unit/integration shaping tests (266 unit tests plus 6,054 shaping cases), and
the interpreter build/test configuration. This is strong regression coverage,
but it does not cover the two failure modes above.

## Part II — HarfBuzz assessment

### 1. Do not port the Rust engine literally

Current HarfBuzz already has `hb_ot_layout_lookup_accelerator_t` and a lazy
`GSUBGPOS::accelerator_t` in `src/hb-ot-layout-gsubgpos.hh`, used by the scan in
`src/hb-ot-layout.cc`. An accelerated subtable already records a typed object
pointer, a compile-time-selected apply function, a digest, and optional external
cache callbacks. LigatureSubst1, PairPos1, PairPos2, and contextual format 2
already build specialized caches. HarfBuzz also recently removed the redundant
per-subtable digest test for one-subtable lookups.

Consequently, duplicating all GSUB/GPOS application semantics into a second
compiled engine would add maintenance risk for less benefit than it did in
Rust: HarfBuzz's packed table overlays already avoid much of the enum/header
construction paid by `read-fonts`. Extend the existing accelerator and external
caches instead. Because an `hb_face_t` owns immutable table blobs, cached
objects can use exact source pointers/offsets; there is no reason to import the
probabilistic content interner.

### 2. GSUB/GPOS work to port, in priority order

1. **Lookup-level next-glyph digest.** Add a conservative second-position
   digest beside the existing union/start digest. Populate it for ligatures and
   coverage-based context where every subtable constrains that position;
   otherwise make it universal. Disable it for reverse and one-component
   ligatures. Test against `buffer->digest` before scanning. This is small,
   independently testable, and had the largest isolated gain in HarfRust.

2. **Contextual rule-set summaries.** Add the first-tested-value summary to the
   external context cache and reject an entire rule set before walking its
   rules. Preserve existing unsafe-to-concat behavior exactly. Do not port
   `RuleFirsts` unless it becomes live and demonstrates a separate win.

3. **PairPos summaries.** Extend PairPos1's cache with a second-glyph summary
   per pair set. Extend PairPos2's cache with the set/digest of nonzero class2
   cells per class1 row. A zero record must continue to have the same matching
   and flag semantics.

4. **Adaptive exact coverage/class indexes and exact lookup reach.** Add
   membership-only range/bitmap/range-list/sorted variants for scanning and
   ranked variants only where parallel arrays require a coverage index. Add
   dense8/dense16/range class maps. Select the representation outside the glyph
   loop. Cache by exact table address/offset and always fall back on allocation
   failure.

5. **Ordered glyph-to-subtable dispatch.** For lookups with several selective
   subtables, build a capped glyph-to-ordered-row map using HarfRust's density
   thresholds as initial values. Measure this independently; recorded gains
   were workload-dependent and smaller than the rejection summaries.

6. **Pass known membership/rank into hot formats.** Once profiles show the
   remaining probes, let apply functions consume the membership or coverage
   index already established by the scanner. Avoid creating a wholesale second
   implementation of all 13 formats unless profiles justify it.

Each step must retain the current path under allocation/cache failure and under
`HB_OPTIMIZE_SIZE` / `HB_NO_OT_LAYOUT_LOOKUP_CACHE`. It should be a separate PR
with isolated benchmarks and serialized-buffer differential tests.

### 3. Non-lookup work for HarfBuzz

[PR #6231](https://github.com/harfbuzz/harfbuzz/pull/6231), “OT: Fuse
redundant per-glyph shaping passes,” merged as `e7fc9442841a` on 2026-09-04.
It retained one commit per optimization:

- **Mask + Unicode-property pass: merged** as `a60b48ef09d4`. Each glyph now
  receives the global mask while its Unicode properties are computed, removing
  `reset_masks()` and its separate buffer traversal. The ZWJ lookahead glyph is
  initialized on the path that advances over it.
- **GDEF properties + digest: merged** as `eed256c0d6f2`. The normal OpenType
  path now accumulates the buffer digest while assigning GDEF glyph properties.
  The AAT/`morx` path retains its post-substitution digest rebuild, and fallback
  glyph-class synthesis remains in the same semantic order.
- **Position initialization: not included.** This remains potentially useful
  but is not a mechanical port.
  HarfBuzz's generic font callbacks may write only requested advance fields, so
  the full-buffer clear remains necessary there. Add an internal built-in-OT
  batch path that overwrites complete records, or distinguish built-in from
  custom callbacks exactly as HarfRust does; keep vertical behavior safe.
- **Fast glyph-map copy: no port needed.** HarfBuzz's
  `hb_ot_map_glyphs_fast` maps in one loop and does not have the Rust
  `out_info` copy to remove.

The merged PR built successfully and passed the buffer and shape API tests,
800 AOTS shaping cases, and 7,272 in-house shaping cases. Performance was left
for independent measurement, aided by the separate commits.

### 4. HarfBuzz acceptance bar

Require byte-for-byte-equivalent glyphs, clusters, positions, and buffer flags;
sanitizer/fuzzer coverage for malformed offsets and nested lookup failure; safe
concurrent first use; graceful OOM/cache-disable fallback; and measurements of
warm throughput, first shape, RSS/cache bytes, and binary size. Benchmark each
layer independently, then together, across Latin, Indic, Arabic/Nastaliq,
variable fonts, mark-heavy fonts, mixed directions/features, desktop/mobile,
and at least one constrained-size build.

The likely HarfBuzz payoff is real, especially from early lookup rejection and
context/pair summaries, but it will be smaller than a direct comparison with
HarfRust's old interpreter because HarfBuzz already performs several of the
same specializations.

dfrg:
Thanks — this is a genuinely useful review. I checked every substantive
claim against the code rather than taking them on faith, and **all six are
correct**. One is worse than reported. Everything below is fixed and pushed
to `perf/compile-more`.

## The benchmark feature bug invalidated my own evidence

This is the most important item and it deserves more than the "configuration
bug" framing.

The self dev-dependency unified features across *tests and examples*, not
just benches, so `--no-default-features --features std` silently built with
`compile-path` on. A probe test prints `compile-path active: true` in that
configuration.

That means the differential verification behind every "identical glyphs and
positions" claim on this branch was invalid: the "interpreted" binary had
the compiled path enabled, so it was comparing the compiled path against
itself.

Fixed (`default-features = false`), rebuilt a genuinely interpreted binary,
re-ran: **99 font/text pairs, 0 differ**. The conclusion survives, but it
was unearned until you found it.

## Blockers

**1. Interner equality — fixed.** Keyed by the table's *position* (which
layout table, how far into it), which is exact by construction rather than
by argument, and recovering the offset now has a test pinning it since it
rests on `read-fonts` behaviour rather than OpenType.

Position alone was the whole change for about an hour, and it was wrong.
It loses sharing when two tables at different offsets hold the same bytes,
and my first measurement said that never happens — because it sampled only
the coverages the *direct* formats name. The contextual formats are where
the duplication is, and Nastaliq is almost entirely contextual: **299
positions naming 177 distinct tables**, and position-only keying grew its
interner from 46.8 KiB to 70.3 KiB.

So the bucket is still chosen by content, and an entry whose position is new
but whose value equals one already in its bucket gets a second key onto the
same `Arc`. Deciding that means building the value to compare — once per
distinct position, not once per reference (a few hundred times against
Nastaliq's sixteen thousand). Comparing compiled values rather than bytes is
the better question anyway: two coverages describing the same glyphs are
interchangeable however they were spelled.

Result: exact key, sharing unchanged at 46.8 KiB, entry still 24 bytes,
0.994 on time.

**2. Nested compile failure — fixed, by removing the disagreement rather
than adding a fallback.** A subtable that cannot be read is now dropped and
the rest of the lookup keeps working; a lookup left with nothing is an empty
slot; an empty slot applies nothing, wherever it is reached from. Nothing
reads the font for a lookup that would not compile.

This rests on something worth stating: every failure path in the compiler is
a malformed read. There is no arm that declines a format it simply does not
handle — all thirteen compile — so a dropped subtable is always one no
reader could have applied, which is where HarfBuzz arrives by declining to
sanitize it. Two regression tests build a GSUB whose only subtable sits past
the end of the table and pin that it leaves an empty slot.

**3. `UNSAFE_TO_CONCAT` — your call, with one correction to the framing.**

The direction matters. HarfBuzz's extra marks are *false positives*: its
filter is a three-word digest, and when the digest admits a second glyph
that cannot actually ligate, it enters the loop and records a hazard no
ligature could have caused. Our set is exact, so we drop exactly those. The
second glyph is already determined when we reject, so appending text cannot
change it and no ligature can become possible — dropping the mark is sound
rather than merely narrower.

I'd still not make that call unilaterally, since it is API output. Two
options:

- **Keep it.** More precise, sound by the argument above, documented at
  `gsub::at_ligature` where it happens.
- **Match bit-for-bit.** One-line switch: use a digest of the same
  second-component set for the reject decision instead of the exact set. The
  awkwardness is that HarfBuzz's digest is `full()` when the subtable has no
  external cache, so "exact compatibility" means reproducing a caching
  artifact — the flag's value depends on whether a cache happened to exist.

Happy to implement either.

## Cleanup — all confirmed dead, all removed

- **`RuleFirsts`** — built, stored, accounted for, never read. Nastaliq's
  reached-lookup memory 2.59× → 2.43×.
- **`LengthEffect`** — the multiple-substitution sequence scan fed a value
  nothing consulted. Enum, join and scan gone.
- **MarkTo metadata** — `bases`, both array offsets, class count, all unread
  because the format applies straight from the font. Dropping the interned
  coverage is most of the win below.
- **`Detail::Lean` doc** — fixed; it described a pair filter removed earlier.
- **Module comments** — swept. Several described designs the code no longer
  has, several narrated what had been tried before arriving here. What
  justifies the current shape stays, in fewer words.

Also `Vec<Subtable>` → `Box<[Subtable]>` (`CompiledLookup` 200 → 104 bytes
across this and earlier cleanups), and `would_apply` is now answered from the
compiled lookup, so `get_lookup` has no reachable caller in a compiled build
— the interpreted cache is allocated and never filled.

Memory, every lookup in the font compiled:

| font | before | after | ratio |
|---|---|---|---|
| Roboto | 55.4 KiB | 44.5 | 1.56× → 1.26× |
| Amiri | 301.0 | 253.8 | 1.50× → 1.27× |
| Nastaliq | 146.3 | 112.4 | 1.66× → 1.27× |
| SourceSerif | 40.4 | 35.4 | 1.55× → 1.36× |
| NotoSans Devanagari | 261.5 | 197.5 | 0.61× → 0.46× |

## Default-on

Reverted; `--features compile-path` again. The two blockers are fixed, but
the rest of your landing sequence isn't done — no fuzzing over malformed and
adversarial fonts, no cold first-shape measurement, no code-size or
constrained-build numbers. Both configurations are tested.

## On methodology

Correct that it isn't in the branch, and that's my fault. One clarification:
"ten short samples" describes the committed criterion bench. The reported
figures come from a harness that scales samples to a fixed 400 ms and takes
medians of interleaved runs — precisely because fixed-count sampling gave a
short-text case 5–8% spread against 1% for a long document. That harness and
the A/B script are local and should be committed so the numbers are
reproducible; I'll do that next.

Also agreed on splitting the four non-lookup fusions into their own PR with
their own evidence, and on the HarfBuzz assessment — porting the rejection
summaries incrementally is the right shape, and #6231 landing two of the
fusions already is good news.

behdad:
Thanks!

Okay I went ahead and stole some of this work into HarfBuzz already:

- The two fused loops showed a small measurable speedup in HB so I landed. https://github.com/harfbuzz/harfbuzz/pull/6231

- Adding a "seconds" digest was a HUGE win in Roboto (~25%). Mostly because it allows for skipping the very heavy ligature lookup that simply ligates latin letters with combining marks into their Unicode precomposed forms. So I landed that in HB as well. https://github.com/harfbuzz/harfbuzz/pull/6232

- That left Roboto shaping be super PairPos-heavy. So I also added the per-PairSet seconds-digest, and that sped up Roboto another 10%. I accidentally pushed that to main instead of PR'ing. This one has some font-dependent memory cost, around 20kb for Roboto. https://github.com/harfbuzz/harfbuzz/commit/8fa0d5da39ab165c51f9d1cbf035f411de909130

- I added a ClassDef cache to input/lookahead of ChainContextFormat2. That sped up Gulzar by like 35%. https://github.com/harfbuzz/harfbuzz/pull/6233

- Add per-RuleSet class digest. This helped NotoNastaliqUrdu by 12%. https://github.com/harfbuzz/harfbuzz/pull/6234

I then had agent port these all into HR: https://github.com/harfbuzz/harfrust/pull/464

image

behdad:
> For analysis and comparison, I can port just these three to HR and measure. I think that's a worthwhile experiment by itself.

https://github.com/harfbuzz/harfrust/pull/464

image

PR #465, opened 2026-09-04
Report unsafe-to-concat consistently in PairPos1 and ligature lookups

dfrg:
Both formats have a summary that lets a lookup conclude "this cannot match" without walking the font, and in both the summary reached that conclusion without reporting the hazard the walk reports. The same font and text then produce different flags depending on which path ran.

GPOS PairPosFormat1

A missed binary search now reports a hazard, which matches the pair-set digest above it and matches HarfBuzz: PairSet::apply reports one on a miss (PairSet.hh:158). A found record that moves nothing no longer reports one -- HarfBuzz does that in format 2, which has an explicit `boring:` label for the case, and not in format 1. The two were swapped here.

Verified against hb-shape on raleway, DIN Alternate, Helvetica and seac, four fonts with format 1 kerning where this crate differed; it now matches HarfBuzz glyph for glyph and flag for flag. No font under tests/fonts distinguishes the change, so there is no test for it.

GSUB LigatureSubstFormat1

A rejecting `seconds` digest now reports the hazard the ligature walk reports. LigatureSet::apply's two paths disagreed on 2 of 399 corpus fonts; they agree on all 399 now.

HarfBuzz disagrees, and looks to have the same inconsistency: its fast path accumulates unsafe_to_concat for exactly this condition (LigatureSet.hh:137), its slow path -- taken when a set holds four or fewer ligatures -- has no such accumulation, and hb-shape reports no hazard on the two fonts the test covers. Neither 14.1.0 nor main has touched these files, so this is not a version skew. This follows the walk, on the grounds that it is the path that inspects the ligatures, but it is worth confirming upstream before anyone relies on it.


dfrg:
> HarfBuzz disagrees, and looks to have the same inconsistency: its fast path accumulates unsafe_to_concat for exactly this condition (LigatureSet.hh:137), its slow path -- taken when a set holds four or fewer ligatures -- has no such accumulation, and hb-shape reports no hazard on the two fonts the test covers. Neither 14.1.0 nor main has touched these files, so this is not a version skew. This follows the walk, on the grounds that it is the path that inspects the ligatures, but it is worth confirming upstream before anyone relies on it.

@behdad this might be worth looking at in HB

behdad:
Codex commenting: I reviewed this against the `UNSAFE_TO_CONCAT` contract and current HarfBuzz. The ligature change is correct, and HarfBuzz needs the equivalent fix, but I think the PairPos change needs one adjustment before merging.

For ligatures, a rejecting `seconds` digest proves that the next glyph cannot be the second component of any ligature. The ordinary walk records that mismatch as unsafe-to-concat; the digest return currently bypasses it. This is a real false negative: changing the nonmatching second glyph to a matching component can form a ligature and change the glyph on the other side of the edit boundary. HarfBuzz's `LigatureSet::apply` has the same omission at its digest return. The analogous contextual RuleSet digest already calls `unsafe_to_concat` before returning, which supports this fix.

For PairPos1, adding `unsafe_to_concat` after a binary-search miss is also correct. However, removing the existing `boring` hazard for a found record whose values are all zero is not correct, even though that matches current HarfBuzz.

A minimal counterexample is:

```fea
pos A X <0 0 0 0>;
pos A V <0 0 -100 0>;
```

Shape `BAXB`, then edit only `X` to `V`. The edit changes `A`'s advance, so the original `A X` range must be unsafe-to-concat. With a synthetic font, current HarfRust marks `A` and `X`; this PR and current HarfBuzz leave both unmarked. An incremental reshaper can consequently start at `X`, reuse `A`, and get the wrong position.

I suggest retaining the existing `boring`/zero-value marking and only adding the missing marking after the binary-search miss. HarfBuzz's `PairSet::apply` appears to have this zero-record bug as well, so upstream needs two fixes: the ligature digest rejection and PairPos1's found-but-zero record.

Minor: the PR text says the ligature slow path is used for four or fewer ligatures, but both HarfBuzz 14.1.0 and current main use one or fewer. Also, `if !ligatures.is_empty()` is redundant in HarfRust because that fast branch is reachable only when `ligatures.len() > 1`.

I ran `cargo test --workspace --all-features` on the PR; all tests pass. I would add a small PairPos zero-record regression test because the existing suite does not catch this false negative.

behdad:
I pushed codex's fix. Please check. If all good, I'll push to HB.

dfrg:
Thanks — this is right, and the zero-record point is a genuine false negative I introduced. I've reset the branch onto @behdad's `d6861eb`.

The argument I should have applied is the symmetry one. A missed search and a found-but-zero record are the same event: the neighbour was consulted and nothing moved. If one deserves the flag, so does the other. HarfBuzz marks the miss and not the zero; HarfRust marked the zero and not the miss. I "fixed" that by matching HarfBuzz, which meant deleting the half that was already correct. The union is the only coherent position, and the `pos A X <0 0 0 0>` / `pos A V <0 0 -100 0>` counterexample makes the cost concrete.

I confirmed both upstream bugs against `724ef405f`:

- `LigatureSet::apply` line 138, `if (seconds && !seconds->may_have (second)) return_trace (false);` — returns with no `unsafe_to_concat`.
- `PairSet::apply` — `unsafe_to_break` only under `applied_first || applied_second`, no `else`.

## Two corrections to my PR text

**The `<= 4` threshold is wrong.** 14.1.0 and current main both use `num_ligs <= 1`. The `<= 4` I quoted was a transient state at the stale commit my checkout was sitting on; `cf614fc85 [LigatureSet] Use slow path only if <= 1 ligatures` restored it.

**"Neither 14.1.0 nor main has touched these files" is wrong, and wrongly derived.** I ran `git log 14.1.0..HEAD -- <files>`, got no output, and read that as "unchanged". My HEAD predated the tag, so the range was simply empty. I should have checked ancestry before concluding anything from it.

The second error matters beyond the wording, because my *explanation* of the ligature evidence was wrong even though the change was right. 14.1.0 has `num_ligs <= 1` **and** already has the `seconds` digest, so on `gsub4_1_multiple_ligatures_f1/f2` HarfBuzz took the fast path and returned at the digest check. It was never "the slow path under-reports" as the commit message claims — it is the digest early-return, i.e. exactly the omission being fixed. That is a cleaner story and strengthens the ligature change rather than qualifying it.

Also, "matches HarfBuzz glyph for glyph and flag for flag" on raleway / DIN Alternate / Helvetica / seac now needs qualifying: those fonts exercise the search-miss path, where we do match. On zero-valued records we now deliberately mark more than HarfBuzz does, pending the upstream fix.

## On the missing test

Before seeing `d6861eb` I scanned all of `tests/fonts` for a PairPosFormat1 record with all-zero value records and found **none**, which independently confirms "the existing suite does not catch this false negative" — a synthetic font was the only way to pin it down.

## One follow-up

`perf/compile-more` carries a copy of this PairPos change in its compiled path (`compile::gpos::pair_done`, via a `still_is_hazard` flag that I set to `false` for format 1). It needs the same zero-record restoration, and its compiled-vs-interpreted differential will hold it honest once both sides agree. I'll handle that on that branch.

dfrg:
We have agreement on the fix. Going to merge and update the compile branch to match behavior.

behdad:
Thanks. I'll port to HB.

PR #475, opened 2026-09-06
[aat] Cache resolved morx subtable views

behdad:
## Summary

- resolve `read-fonts` `morx` subtable views once while preparing AAT tables
- reuse the borrowed views during shaping instead of reconstructing them after every subtable screen
- retain the existing reconstruction path for malformed or unavailable cached views
- keep contextual views covariant so the cache does not restrict `hb_font_t` lifetimes

No font-table arrays are decoded or copied. On x86_64, the cache is 112 bytes per `morx` subtable: about 4.4 KiB for Devanagari Sangam MN. The stripped library grows by about 11 KiB.

This is substantially smaller than the previously reverted compiled-state-machine cache, which used 88--152 KiB for the tested fonts.

## Performance

Measured shaping-time changes against current `main`, with the same optimized Fontations dependency:

| Workload | Change |
| --- | ---: |
| Devanagari Sangam MN, Hindi words | about -4.5% |
| Geeza Pro, Persian words | about -2.7% |
| Lucida Grande, English text | about -3.1% |
| Menlo, English text | about -2.3% |

For Devanagari Sangam MN, hardware counters also showed about 3.9% fewer instructions, 5.5% fewer branches, and 6.4% fewer branch misses.

## Testing

- `cargo fmt --all -- --check`
- `cargo build --no-default-features --features=libm`
- `cargo clippy --all-features --all-targets -- -D warnings`
- `cargo test --workspace`

### DIFF
--- a/harfrust/src/hb/aat/layout_morx_table.rs
+++ b/harfrust/src/hb/aat/layout_morx_table.rs
@@ -12,7 +12,7 @@ use alloc::{vec, vec::Vec};
 use read_fonts::tables::aat::{self, ExtendedStateTable, NoPayload, StateEntry, StateTable};
 use read_fonts::tables::{mort, morx};
 use read_fonts::types::{BigEndian, FixedSize, GlyphId, GlyphId16};
-use read_fonts::{FontData, ReadError};
+use read_fonts::{FontData, FontRead, ReadError};
 
 trait MorphChain {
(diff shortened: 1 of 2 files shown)


dfrg:
Would be nice if this was viable but any cache that holds a lifetime won’t persist across shaping calls for any real users.

PR #478, opened 2026-09-17
Bump to read-fonts 0.44

akx:
Bumps read-fonts from 0.43.3 to 0.44.0 and fixes what needs to be fixed.

`cargo test` runs on my machine aside from some flakiness around `NULL_FUNC_DROPS` (52dc6939bd816235c850941c4607b414b8b2f0fc). A couple re-runs and it was green.


akx:
Ah... looks like yoke-derive 0.8.3 (published yesterday) isn't compatible with Rust 1.85?

<details>

```
$ cargo +1.85 build
error[E0599]: no function or associated item named `from_utf8` found for type `str` in the current scope
   --> yoke-derive-0.8.3/src/lib.rs:202:32
    |
202 |         let underscores = str::from_utf8(&underscores).expect("_ is ASCII and thus UTF-8");
    |                                ^^^^^^^^^ function or associated item not found in `str`
    |
help: you are looking for the module in `std`, not the primitive type
    |
202 |         let underscores = std::str::from_utf8(&underscores).expect("_ is ASCII and thus UTF-8");
    |                           +++++
```

</details>

**EDIT:** Fixed in the second commit. `cargo +1.85 build` and `cargo build -p harfrust --target thumbv7em-none-eabihf --no-default-features --features=libm` work locally too.

akx:
Fixed the "multiple versions for dependency `syn`: 2.0.119, 3.0.6" lint issue too (by allowing both syn 2 and syn 3 in clippy.toml).

dfrg:
Thanks! This looks good, but I think we want to hold off on read-fonts updates for just a bit. Releasing will require a breaking change and there are updates coming to read-fonts soon (hopefully next week) that will provide real performance and correctness improvements.

akx:
@dfrg Okiedoke, thanks for the heads-up! Looking forward to it.

